PR #79, opened 2014-06-27
Class for handling arrays of patches

DanielCardonaRojas:
This class inherits form PdFile class and makes creates a NSMutableArray to which patches may be added at any time. It also lets you communicate with patches independently throw and array index, grouped by name or sending the same message to all. Implementation of list messages and single messages is not finished.

### DIFF
--- a/objc/PdPatchArray.h
+++ b/objc/PdPatchArray.h
@@ -1,0 +1,70 @@
+//
+//  PdPatch.h
+//  GuitarTuner
+//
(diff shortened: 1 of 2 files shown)


danomatika:
I agree. This is a good implementation and useful code, but we want to keep libpd basic and users can arrange patch access and control themselves. So far we are keeping away from language wrapper specific convenience classes and functions beyond providing a basic sound layer (iOS and Android).

PR #80, opened 2014-08-15
Allow tables to be resized and accessed as FloatBuffers

mhroth:
These functions expose additional functionality to the Java domain. Tables may be resized. And instead of reading and writing to them via a Java-array intermediary, they may now be accessed directly as `java.nio.FloatBuffer`s. Finally, the ability to execute a generic `Runnable` in a thread-safe way has been added, ensuring that the audio thread is not running.

### DIFF
--- a/java/org/puredata/core/PdBase.java
+++ b/java/org/puredata/core/PdBase.java
@@ -111,8 +111,11 @@ public final class PdBase {
    */
   public static void executeSynchronizedRunnable(Runnable r) {
     lockNative();
-    r.run();
-    unlockNative();
+    try {
+      r.run();
+    } finally {
+      unlockNative();
+    }
   }
 
   /**


danomatika:
@nettoyeurny What's your take on this now that locking is built into the C core? Access to filling arrays directly would solve the problem of missing native file format support in libpd for iOS, etc.

nettoyeurny:
I have to say I'm still a bit worried about this idea because a large
transfer might cause the lock to be held for a long time. What about the
idea of swapping arrays, so that read/write operations can be done without
the lock?

danomatika:
I agree with Peter regarding the direct buffer access. It's probably best done manually outside of libpd when you really need it. Also, newer capabilities such as integrating double support internally would mean we couldn't guarantee the buffer type you are accessing.

However, I think being able to resize an array is definitely useful and ported your implementation to #281.

danomatika:
>What about the idea of swapping arrays, so that read/write operations can be done without the lock?

I feel this takes use more into wrapper territory. The base C api should be simple and straight forward. I could see something a little more complicated like this being introduced as an additional helper utility, etc.

Issue #77, opened 2014-06-16
Ocassional crashes on libpd_queued_receive_pd_messages


rvega:
Hi. 
I've been using the cpp-queued branch of libpd within a QT C++ application compiled with clang in MacOS. 

To get messages from PD, I created a receiver class based on the one in the cpp folder, and to actually receive the messages, I am calling PdBase::receiveMessages() from the callback of a [QTTimer](http://qt-project.org/doc/qt-5/qtimer.html) (which runs in the main thread of the app).

Every once in a while (I have to sit and wait for a few minutes) the app will crash. I ran the app within lldb (clang's debugger) a few times and I found weird behavior in libpd_queued_receive_pd_messages every time before the crash. The stack trace:

```
(lldb) thread backtrace
* thread #1: tid = 0x726c, 0x00007fff8cf88670 libsystem_c.dylib`strlen + 16, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00007fff8cf88670 libsystem_c.dylib`strlen + 16
    frame #1: 0x0000000100031e65 Bounces`pd::PdBase::PdContext::_bang(char const*) [inlined] std::__1::char_traits<char>::length(__s=0x0000000000000000) + 8 at string:651
    frame #2: 0x0000000100031e5d Bounces`pd::PdBase::PdContext::_bang(char const*) [inlined] basic_string(this=0x00000001000c0df0) at string:1866
    frame #3: 0x0000000100031e5d Bounces`pd::PdBase::PdContext::_bang(char const*) [inlined] basic_string(this=0x00000001000c0df0) at string:1870
    frame #4: 0x0000000100031e5d Bounces`pd::PdBase::PdContext::_bang(source=0x0000000000000000) + 45 at PdBase.cpp:999
    frame #5: 0x00000001000c15a7 Bounces`libpd_queued_receive_pd_messages [inlined] receive_bang + 18 at z_queued.c:68
  * frame #6: 0x00000001000c1595 Bounces`libpd_queued_receive_pd_messages + 133 at z_queued.c:343
    frame #7: 0x00000001000388b3 Bounces`DBM::PdBridge::qt_static_metacall(_o=0x000000010521d2c0, _c=<unavailable>, _id=<unavailable>, _a=<unavailable>) + 1491 at moc_PdBridge.cpp:148

...

[...]

Issue #43, opened 2013-09-11
EXC_BAD_ACCESS in Obj-C PdBase listHook


danomatika:
In the Obj-C wrapper on iOS, I'm running a patch that is sending lots of continual list data to libpd. Over time, I keep seeing an EXC_BAD_ACCESS on 

    NSString *source = [[NSString alloc] initWithCString:src encoding:NSASCIIStringEncoding];

in https://github.com/libpd/libpd/blob/master/objc/PdBase.m#L120

The debugger shows the src cstring is not null so I'm not sure what's going on ... maybe it's not being null-terminated?

nettoyeurny:
Should be fixed in Version 0.8.1 (now pushed to master branch).

PR #95, opened 2015-06-06
Fix circular buffer read/write issue for compound PD messages.

monkeyswarm:
A fix for 
https://github.com/libpd/pd-for-android/issues/4
in which bad pd messages are sent out, by reading the buffer in between the multiple buffer writes of a compound pd messages (e.g. list, message).

and likely related to
https://github.com/libpd/libpd/issues/77
https://github.com/libpd/libpd/issues/43

### DIFF
--- a/libpd_wrapper/util/ringbuffer.c
+++ b/libpd_wrapper/util/ringbuffer.c
@@ -21,6 +21,7 @@ ring_buffer *rb_create(int size) {
   }
   buffer->size = size;
   buffer->write_idx = 0;
+  buffer->write_readable_idx = 0;
   buffer->read_idx = 0;
   return buffer;
 }
@@ -46,8 +47,9 @@ int rb_available_to_write(ring_buffer *buffer) {
 int rb_available_to_read(ring_buffer *buffer) {
   if (buffer) {
     int read_idx = __sync_fetch_and_or(&(buffer->read_idx), 0);
-    int write_idx = __sync_fetch_and_or(&(buffer->write_idx), 0);
-    return (buffer->size + write_idx - read_idx) % buffer->size;
+    int write_readable_idx =
+         __sync_fetch_and_or(&(buffer->write_readable_idx), 0);
+    return (buffer->size + write_readable_idx - read_idx) % buffer->size;
   } else {
     return 0;
   }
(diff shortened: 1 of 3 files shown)


danomatika:
Pinging @nettoyeurny & @chr15m

chr15m:
The code itself looks fine to me. (I don't know much about the atomics so take my input with a grain of salt.)

My one comment would be to rebase the two commits into one - the first commit only changes indentation - and to try to make the indentation match the original so that the change itself is a minimal patch just affecting the lines that are actually changing.

Would like to hear from @nettoyeurny before we merge but this looks fine to me. Thanks @monkeyswarm!

monkeyswarm:
Sure thing, looks like my formatter got a little overzealous there. New commit is a squash of the second commit, with indentation fixed.

As for the atomics, I too am a little fuzzy on whether an atomic read/write is in order for the usage of buffer->is_in_write_group (ringbuffer.c lines 71, 100, and 104). Given that the value is only zero or one, I don't think a torn write/read would have an effect. I'm happy to add the atomics if needed and/or it adds peace of mind (or cross-hardware/OS predictability). Thanks!

nettoyeurny:
Sorry about the tardy reply...

My take on this is that the ring buffer should have a write method that takes multiple buffers and updates the indices only once. Possible implementations would involve varargs or a second write method that takes a pair of buffers.

Any thoughts would be appreciated. I'll mull this over and get back to you.

nettoyeurny:
Alright, I've thought about this and I think that varargs are the way to go.

A tentative fix is here: https://github.com/nettoyeurny/libpd/commit/96d0a4479868effd3888a0c28eda0e29a2324411

Warning: I haven't yet had a chance to really test this; I just wanted to send you something as quickly as possible because I've been awol for too long. Please let me know what you think.

While working on this, I noticed some bit rot in the Java branch. Since the Java branch has better test coverage, I'll fix that in a separate step and then revisit this issue.

nettoyeurny:
Java tests are fixed, and they pass with the varargs revision. I feel ready to push this change unless you have objections. Please let me know!

monkeyswarm:
Looks good to me.

chr15m:
:+1:

nettoyeurny:
Quick update: I've been in touch with @danomatika and we're all in agreement. I just pushed the varargs change to master. I'll close this pull request. Thanks, everybody!

PR #96, opened 2015-06-19
start (and stop) Pd GUI from libpd

millerpuckette:
introduces functions libpd_start_gui() and libpd_stop_gui().  Requires corresponding updates to Pd
source.

Also compile 'fixes' for MSVC which hates declarations that aren't at heads of blocks; and
fixes a problem if someone calls midi_get_APIs() (which gets done when GUI is started).

### DIFF
--- a/libpd_wrapper/s_libpdmidi.c
+++ b/libpd_wrapper/s_libpdmidi.c
@@ -52,7 +52,10 @@ void outmidi_byte(int port, int value) {
 }
 
 // The rest is not relevant to libpd.
-void sys_get_midi_apis(char *buf) {}
+void sys_get_midi_apis(char *buf)
+{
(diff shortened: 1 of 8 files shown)


danomatika:
Thanks. I just created a `gui_updates` branch. Can you close this PR and resubmit to it instead of `master` so we can compartmentalize these changes?

richardeakin:
Super cool, especially since now all of pd is available via the submodule so why not expose an optional GUI.

I'm also excited to try building with Visual Studio again, that was the holdup on getting [Cinder-PureDataNode](https://github.com/notlion/Cinder-PureDataNode) to work on windows..

millerpuckette:
closing and re-submitting to gui_updates branch

PR #97, opened 2015-06-19
opening and closing Pd GUI from libpd

millerpuckette:
introduces functions libpd_start_gui() and libpd_stop_gui(). Requires corresponding updates to Pd
source.

Also compile 'fixes' for MSVC which hates declarations that aren't at heads of blocks; and
fixes a problem if someone calls midi_get_APIs() (which gets done when GUI is started).

### DIFF
--- a/libpd_wrapper/s_libpdmidi.c
+++ b/libpd_wrapper/s_libpdmidi.c
@@ -52,7 +52,10 @@ void outmidi_byte(int port, int value) {
 }
 
 // The rest is not relevant to libpd.
-void sys_get_midi_apis(char *buf) {}
+void sys_get_midi_apis(char *buf)
+{
+    strcpy(buf, "{}");
+}
 void sys_listmididevs(void) {}
 void sys_get_midi_params(int *pnmidiindev, int *pmidiindev,
     int *pnmidioutdev, int *pmidioutdev) {}
--- a/libpd_wrapper/util/ringbuffer.c
+++ b/libpd_wrapper/util/ringbuffer.c
@@ -12,8 +12,9 @@
 #include <string.h>
 
 ring_buffer *rb_create(int size) {
+  ring_buffer *buffer;
   if (size & 0xff) return NULL;  // size must be a multiple of 256
-  ring_buffer *buffer = malloc(sizeof(ring_buffer));
+  buffer = malloc(sizeof(ring_buffer));
   if (!buffer) return NULL;
   buffer->buf_ptr = calloc(size, sizeof(char));
   if (!buffer->buf_ptr) {
@@ -55,12 +56,12 @@ int rb_available_to_read(ring_buffer *buffer) {
 }
 
 int rb_write_to_buffer(ring_buffer *buffer, int n, ...) {
+  int write_idx, available, i;
   if (!buffer) return -1;
-  int write_idx = buffer->write_idx;  // No need for sync in writer thread.
-  int available = rb_available_to_write(buffer);
+  write_idx = buffer->write_idx;  // No need for sync in writer thread.
+  available = rb_available_to_write(buffer);
   va_list args;
   va_start(args, n);
-  int i;
   for (i = 0; i < n; ++i) {
     const char* src = va_arg(args, const char*);
     int len = va_arg(args, int);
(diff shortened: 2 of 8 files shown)


danomatika:
Thanks. @nettoyeurny is going to look in as well.

danomatika:
Also, I updated the pure-data repo from upstream master so that commit should checkout correctly.

millerpuckette:
Cool - I didn't know if that worked automatically or not, I guess not :)
I'll probably be finding other fixes to Pd (haven't tested GUIs in the
case one has multiple Pd instances around and I bet it won't work at all)

As I update Pd in the future, I'm not sure whether I can re-sync the fork and
generate a pull request to re-sync it on the official git, or whether that
works through a different mechanism.  Will try to figure that out for next
time.

cheers
Miller

danomatika:
The pure-data sources are a submodule that tracks pure-data/puredata on Github which itself tracks the sourceforge git. If work goes into the sourceforge git (upsrteam), we can pull it into pure-data at anytime. I created pure-data/pure-data mainly for having a place to work on things and generate patches to send on to you and in case we needed to test something libpd only.

--------
Dan Wilcox
@danomatika <https://twitter.com/danomatika>
danomatika.com <http://danomatika.com/>
robotcowboy.com <http://robotcowboy.com/>

chr15m:
I've invited Miller to the pure-data organisation as an owner. We could pin libpd to something like pure-data/libpd-pure-data and keep pure-data/pure-data as an exact mirror of the original so as not to confuse people.

danomatika:
pure-data/pure-data is already a mirror from sourceforge

chr15m:
It's not up to date because we only pull it at major Pd version releases. The URL pure-data/pure-data implies to users that it is some kind of canonical source of Pd, which at the moment it isn't.

Maybe this isn't even a problem, in which case, sorry for the noise!

danomatika:
One of us running fetch upstream and then a merge is not rocket science.

Well, if you remember I originally used the sourceforge git in the submodule but there was some instance of needing to try a patch etc so it made more sense to keep a local mirror in case we needed to modify pure-data and send the changes upstream. If, now, we don't need to be able to do that then, by all means, I can change the remote back however we lose the nice ability to browse the source via GIthub's UI.

chr15m:
Thinking about it some more, I guess what I am asking is if @millerpuckette decided at some point to push directly to Github pure-data/pure-data instead of [or in addition to] Sourceforge then we would be ok with that, correct?

I think if the situation arose where we definitely needed libpd to be pinned to some alternate version of Pd for some reason, then we could create a new branch or repository to do that.

You're right that it is not difficult to fetch upstream and merge as we are doing now.

millerpuckette:
OK... sounds like I should start pushing to the github Pd project and
let you guys run the libpd repo in whatever way is easiest.  It's moot
for the moment since I won't be able to do much work on Pd for a month
or so (big project coming to a head Aug. 6 :)

But then I want to make some changes that might temporarily break 
stuff...
Miller

danomatika:
No, that's not what I'm saying at all. It was only because we thought we might need to do libpd specific stuff. That hasn't actually been the case since you've basically handled all the stuff I've found over the last year or so. I'm going to change the remote back to source forge.

danomatika:
As a follow up, if we do experiment with major changes, that can just as easily be handled in a branch using a different remote for the submodule. I see that as a possible encase but, so far, the main usage has been straight from the source forge upstream so it probably makes the most sense to cut out the mirror.

danomatika:
Ok, sorry for waffling. Chris you are right. We want things to be as smooth and easy as possible and the mirror was a good idea but has actually been getting in the way. When we do want to do work to submit upstream, we can use the mirror via a diff submodule link in a branch.

I updated the submodule link to point back to source forge. Run the following to get things up to speed (warning, removes & checks out pure-data):

    rm -rf pure-data
    git submodule init
    git submodule update

chr15m:
Ping @nettoyeurny for review (sorry Peter, I know you're very busy!).

@millerpuckette so if you decide to push to github.com/pure-data/pure-data let us know and we'll update the submodule link. Either SF or Github is fine though so totally up to you.

danomatika:
Actually Chris, I already changed the submodule link to sourceforge.

nettoyeurny:
Arrgh! I totally missed this one. Sorry about the very tardy reply --- is this still under consideration?

danomatika:
Yes. We can use your input :)

nettoyeurny:
I'll look into it. (Is there anything else that I should have responded to ages ago?)

danomatika:
Maybe #114. Check the open issues list. Another thought was I've been thinking of transitioning the Makefile to auto tools for a while ...

danomatika:
We should resurrecting this thread!

Sorry it devolved into "which pure-data mirror to use". All of these commits look great...

PR #100, opened 2015-08-10
Fixed some cpplib build issues under Windows/Mingw.

eigenbom:
Symbols weren't exporting for the cpplib under Windows/Mingw, so I patched the Makefile.

Also patched the makefiles for two samples to get them running under Windows/Mingw. 

> gcc version 5.2.0 (Rev3, Built by MSYS2 project)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ else
       -I"$(JAVA_HOME)/include" -I"$(JAVA_HOME)/include/win32"
     MINGW_LDFLAGS = -shared -lws2_32 -lkernel32
     LDFLAGS = $(MINGW_LDFLAGS) -Wl,--output-def=libs/libpd.def \
-      -Wl,--out-implib=libs/libpd.lib
+      -Wl,--out-implib=libs/libpd.lib -Wl,--export-all-symbols
     CSHARP_LDFLAGS = $(MINGW_LDFLAGS) -Wl,--output-def=libs/libpdcsharp.def \
       -Wl,--out-implib=libs/libpdcsharp.lib
     CPP_LDFLAGS = $(LDFLAGS)
--- a/cpp/PdBase.hpp
+++ b/cpp/PdBase.hpp
@@ -24,7 +24,10 @@
 
 // define this to use C++11 std::mutex for locking
 #ifdef LIBPD_USE_STD_MUTEX
-    #if __cplusplus < 201103L
+	#ifdef _WIN32
+		// __cplusplus is badly behaved in windows. just assume we have mutex
+		#include <mutex>
+	#elif __cplusplus < 201103L		
         #warning std::mutex requires C++11
     #else
         #include <mutex>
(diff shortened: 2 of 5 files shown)


danomatika:
Thanks! I'll get to this soon.

cpp/PdBase.hpp
```diff
@@ -24,7 +24,10 @@
 
 // define this to use C++11 std::mutex for locking
 #ifdef LIBPD_USE_STD_MUTEX
-    #if __cplusplus < 201103L
+	#ifdef _WIN32
+		// __cplusplus is badly behaved in windows. just assume we have mutex
```
danomatika: I don;t like this. Why just *assume*? Maybe someone is using libpd with an older C++ library and can't compile it with C++11 support?

Makefile
```diff
@@ -24,7 +24,7 @@ else
       -I"$(JAVA_HOME)/include" -I"$(JAVA_HOME)/include/win32"
     MINGW_LDFLAGS = -shared -lws2_32 -lkernel32
     LDFLAGS = $(MINGW_LDFLAGS) -Wl,--output-def=libs/libpd.def \
-      -Wl,--out-implib=libs/libpd.lib
+      -Wl,--out-implib=libs/libpd.lib -Wl,--export-all-symbols
```
danomatika: Shouldn't this be handled on the line above? `--output-def=libs/libpd.def` Aren't the symbols available in the def file? Just double-checking.

samples/c/pdtest/Makefile
```diff
@@ -9,6 +9,7 @@ else
   ifeq ($(OS), Windows_NT)  # Windows, use Mingw
     SOLIB_EXT = dll
     PDNATIVE_PLATFORM = windows
+    CC = gcc
```
danomatika: If you're running in msys, isn't this set by default?

cpp/PdBase.hpp
eigenbom: The problem is that some versions of msvc implements mutex but have _cplusplus=199711L. I agree it is bad to assume. One way around this would be to use _MSC_VER or a user-configurable header.

Makefile
eigenbom: I can't remember why I had to do this, sorry.

samples/c/pdtest/Makefile
eigenbom: Hmm not sure, maybe it was set to cc by default.

cpp/PdBase.hpp
danomatika: Ok, I follow you. In looking around stack overflow, it seems like doing an explicit check will not always work since compilers can lie and not everything might be supported. In that case, it's better to leave the responsibility on the user and honor the define. I've decided to drop the _cplusplus check.

danomatika:
At this point, I've brought in most of your proposed changes. Can you try the latest libpd master to see of things are working on your system?

eigenbom:
Ok, cheers. I'm not sure if and when I'll be able to test this, probably not for quite some time. Glad you found my patch useful.

danomatika:
Closing due to inactivity. Assuming the updates are working...

PR #104, opened 2015-09-28
Install command should point to libpd official repo

yairsz:
Just a quick PR to point the podfile install command to the official libpd repo instead of a fork.


danomatika:
Ah, whoops. Thank You,

PR #109, opened 2015-10-16
first stab at supporting the win api atomics in the ringbuffer

danomatika:
Adding support for the win api (& Visual Studio) in the ring buffer by using a macro.

From my research, I feel like I've found the equivalent functions between the apis:

* [gcc atomics builtins](http://gcc.gnu.org/onlinedocs/gcc-4.4.3/gcc/Atomic-Builtins.html)
* [win api interlocked variable access](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684122(v=vs.85).aspx)

As I don't have a windows machine, this will obviously need to be verified.

Pinging @nettoyeurny.


danomatika:
@nettoyeurny @residuum Any thoughts?

danomatika:
I also brought in Peter's work from the apple_atomics branch.

libpd_wrapper/util/ringbuffer.c
```diff
@@ -14,6 +14,21 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef __APPLE__ // apple atomics
+# include <libkern/OSAtomic.h>
+# define SYNC_FETCH(ptr, val) OSAtomicOr32Barrier(val, ptr)
+# define SYNC_COMPARE(ptr, oldval, newval) \
```
nettoyeurny: Let's call this SYNC_COMPARE_AND_SWAP since the swap is really the main part here.

libpd_wrapper/util/ringbuffer.c
```diff
@@ -14,6 +14,21 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef __APPLE__ // apple atomics
+# include <libkern/OSAtomic.h>
+# define SYNC_FETCH(ptr, val) OSAtomicOr32Barrier(val, ptr)
```
nettoyeurny: Since val is always 0, let's just get rid of val and call OSAtomicOr32Barrier(0, ptr) directly.

nettoyeurny:
Looks good --- thanks for taking care of this! I added a couple of comments, strictly cosmetic.

danomatika:
Updated

alisomay:
It looks like it has been some years but I've created #350 which might be related to this.

Issue #87, opened 2015-02-13
Unable to build C# API - libpdcsharp.dll


milanjaros:
Hi,
I'm unable to [build the libpd using mingw](https://github.com/libpd/libpd/wiki/Building-the-C%23-Api).

```make clean``` goes well, so it seems mingw is set-up properly, but

```make csharplib``` throws:
make: *** No rule to make target 'pure-data/src/d_arithmetic.o', needed by `libs/libpdcsharp.dll'. Stop.

@tebjan was so kind and prepared how-to but I'm not able to follow it.
Can you, please, suggest anything?

**Comment by danomatika** [comment]
Did you try mingw_build.bat?

**Comment by danomatika** [comment]
I think I know you're problem. You need to checkout the submodule which has the libs sources. See here: https://github.com/libpd/libpd#pure-data

**Comment by milanjaros** [comment]
Great, thanks! 
After cloning "pure-data" sources I've got also:
     `pure-data/src/d_soundfile.c:17:21: fatal error: pthread.h: No such file or directory`

So, I followed [your proposal from another thread](https://github.com/patricksebastien/libpd4unity/issues/4#issuecomment-44594141) and installed MinGW pthreads: 
`mingw-get install mingw32-pthreads-w32-dev`
and I've got:
```
c:\tools\mingw\include\pthread.h:320:8: error: redefinition of 'struct timespec'

 struct timespec {
        ^
In file included from c:\tools\mingw\include\unistd.h:95:0,
                 from pure-data/src/d_soundfile.c:15:
c:\tools\mingw\include\parts\time.h:65:8: note: originally defined here
 struct timespec
        ^
make: *** [pure-data/src/d_soundfile.o] Error 1
```
I'm stuck again.

danomatika:
Dunno. Looks like a conflict between a defined type in both of those headers. In this case it's a MinGW environment problem and not with libpd. You might check the headers to see if there some define that needs to be set to ignore one of the definitions, etc.

milanjaros:
Great, I build it. Sorry for bothering... I've found this is really some [MinGW issue with incompatible libraries](https://lists.gnu.org/archive/html/bug-gnulib/2015-01/msg00042.html).

So, I went to `MinGW\include\pthread.h` line 320 and commented out the scruct. 

Thank you very much for your patience. 

Milan ;)

webfd:
wo yet has a same problem

milanjaros:
Have you tried to modify the pthread.h?

[...]

PR #113, opened 2016-01-26
Bring C# bindings up to date

residuum:
- Use correct calling convetions for native dll, see http://stackoverflow.com/a/3506938/124983
- Update test project
- Make it usable for Mono with Linux


residuum:
I am not sure, why the test project contains code to load dll libraries via kernel32.dll, because that leads to exceptions while running the unit tests on Linux. Removing these lines let the tests pass.

danomatika:
Thanks for the work! The csharp lib definitely needed some love. Let me know when you're done.

residuum:
This is a solution that is working in Windows and Linux without changing the API. So, please merge this. Mac OS X should be working, but I am not able to test it.

It is still a thin wrapper around the C library, and I will think about making it more .NET/Mono-ish, but that will need some more work, especially some experience with audio programming.

tebjan:
the dynamic DLL loading is needed because each test hast to load and unload the libpd library and the test apcth. the [SetUp] and [TearDown] attributed methods are called before and after each test. if it does not load/unload the patch, the static fields of the libpd dll will not be cleared and the patch can have a different state. thus the tests are not objective in this case.

but i think its not that important to have the test run on linux and mono.

tebjan:
also please check the license of libgcc_s_dw2-1.dll, i think its GPL and its not allowed to ship it with libpd. but i don't know exactly how the license policy of this repo is.

tebjan:
currently not sure about the calling convention changes. the stackoverflow thread does not exactly match the libpd situation since libpd does not specify __cdecl on the exported method calls, which means the compiler decides what the convention is. this has to be checked with the defaults of the used compilers, e.g. the one used in the mingw build...

residuum:
OK, I have left out libgcc_s_dw2-1.dll, and used `--static-libgcc` instead.

I have also included a cross-platform way to have the unit tests run, at least on Windows and Linux.

About the calling convention: As long as we only take the MinGW builds, then it should be fine with __cdecl.

Makefile
```diff
@@ -26,7 +26,7 @@ else
     LDFLAGS = $(MINGW_LDFLAGS) -Wl,--output-def=libs/libpd.def \
       -Wl,--out-implib=libs/libpd.lib
     CSHARP_LDFLAGS = $(MINGW_LDFLAGS) -Wl,--output-def=libs/libpdcsharp.def \
-      -Wl,--out-implib=libs/libpdcsharp.lib
+	-static-libgcc -Wl,--out-implib=libs/libpdcsharp.lib
```
residuum: I am not sure if this should be included in LDFLAGS instead for all builds.

tebjan:
nice work on the dynamic loading on linux!
had no time to check the compiler flags, but if it works for you then fine with me.

last words should come from the repo owners...

danomatika:
We defer to you guys since y'all have the csharp knowledge. Let me know when your ready to merge and I'll just do a quick once over and go.

residuum:
I have just squashed all commits into a single one.

residuum:
I could add a simple example on how to use libPd with NAudio, but I won't have time until the weekend. If you merge this before, then I will open a new PR.

tebjan:
ok, I'll do a quick test on my machine (only window though) and report back.

tebjan:
i had two issues:

1.) tests
an error with the tests, the dollarzero was 1003 for me instead of the expected 1002... i don't know what this exactly means, has this changed in libpd internally? @residuum do you get the same error with the dollarzero test?

had to update it like this:
https://github.com/tebjan/libpd/commit/c16c8075abcfa8d0fb938c473ead8e045fa243eb

2.) build
in mingw i had to install mingw32-base meta package and mingw32-pthreads-w32 dev package.
i also had to mod the pthread.h as described here: https://github.com/libpd/libpd/issues/87
the documentation should probably reflect this mingw problem...

residuum:
Hmm, yes I also get 1003 on Windows, but 1009 on Linux. I thought, it was a problem with previously running instances of Pd or libPd that somehow were still hanging around, but that seems not to be the case.

residuum:
The Linux problem comes from a null pointer from dlopen and the execution as the 4th test. It seems, that opening a file from the local directory is not working correctly.

residuum:
I have just checked with Pd: $0 on the first opened patch is now 1003, so it seems that there was a change in the underlying code.

IIRC there is a refactoring going on to make Pd thread-safe, probably this results in yet another call to `canvas_new()` which starts with 1000: https://github.com/pure-data/pure-data/blob/master/src/g_canvas.c#L356

residuum:
I could not load the local file with `dlopen` in Linux for the unit tests, it always return `IntPtr.Zero` (`null` in C). 

I have created a simple example for using libPd with NAudio.

residuum:
OK, I did change that to test for 1003 and re-rebased the commits.

residuum:
ping.

residuum:
ping.

Kyew:
pong

SAUVAGEOT Paul-Arthur.
Software Engineer.

2016-03-24 17:51 GMT+01:00 Thomas Mayer <notifications@github.com>:

residuum:
After testing it with NAudio in Linux, I have discovered, that everything is working, except the buffer after processing the sound in libpd is always constant 0. I will look into it on the weekend.

danomatika:
Cool. If you get a chance, can you also check the current master branch? I updated to the latest pure-data master in anticipation of the upcoming 0.47 release. The main thing is that expr is now part of the pure-data core src folder so I've updated the Makefiles but haven't checked the build on all platforms.

residuum:
I have just rebased my chances on the current master. Works (apart from issue #130), but that can be resolved indepently.

residuum:
Once this is merged, I will make a [NuGet package](https://www.nuget.org/) for LibPD.

danomatika:
Looks good overall.

A few things I noticed:

* The code formatting in `samples/csharp/pdtest/LibPDTests.cs` differs from all the other csharp files
* Maybe it's a good idea to place the sharp / windows libs in platform based subfolders? Aka `libs/csharp` or `libs/mingw32`.

residuum:
> Maybe it's a good idea to place the csharp / windows libs in platform based subfolders? Aka `libs/csharp` or `libs/mingw32`.

What do you mean by that? Do you mean the `libwinpthread-1.dll` versions for 32 and 64 bit systems? In that case I would prefer the `libs/mingw32` and `libs/mingw64` subfolders, as the dll are not specific to C#, but also for other Windows builds.

danomatika:
Yeah, `libs/mingw32` and `libs/mingw64` looks good.

PR #127, opened 2016-03-07
Fix AVAudioSession method call on iOS 5 and below

monkeyswarm:
[AVAudioSession sharedInstance].inputAvailable is on iOS 6 and above, causes crashes on iOS 5. Adding preproccessor directives to use both the old/deprecated and newer method names.


danomatika:
I'm thinking about reworking this anyway to use respondsToSelector: instead of the compile time checks which only check the MINIMUM version. That's the better solution I think.

danomatika:
On second thought, this has been working fine for a while, so why change it now. :)

PR #132, opened 2016-05-31
[WIP] GUI open/close support

danomatika:
Re-adding @millerpuckette prelim GUI control from #97.

### DIFF
--- a/libpd_wrapper/s_libpdmidi.c
+++ b/libpd_wrapper/s_libpdmidi.c
@@ -6,6 +6,7 @@
 #include "s_stuff.h"
 #include "z_libpd.h"
 #include "z_hooks.h"
+#include <string.h>
 
 #define CLAMP(x, low, high) ((x > high) ? high : ((x < low) ? low : x))
 #define CLAMP4BIT(x) CLAMP(x, 0, 0x0f)
@@ -52,7 +53,9 @@ void outmidi_byte(int port, int value) {
 }
 
 // The rest is not relevant to libpd.
-void sys_get_midi_apis(char *buf) {}
+void sys_get_midi_apis(char *buf) {
+  strcpy(buf, "{}");
+}
 void sys_listmididevs(void) {}
 void sys_get_midi_params(int *pnmidiindev, int *pmidiindev,
     int *pnmidioutdev, int *pmidioutdev) {}
--- a/libpd_wrapper/z_libpd.c
+++ b/libpd_wrapper/z_libpd.c
@@ -19,6 +19,7 @@
 #include "s_stuff.h"
 #include "m_imp.h"
 #include "g_all_guis.h"
+#include "g_canvas.h"
 
 #if PD_MINOR_VERSION < 46
 # define HAVE_SCHED_TICK_ARG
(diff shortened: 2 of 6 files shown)


danomatika:
pinging @nettoyeurny @kmatheussen

libpd_wrapper/s_libpdmidi.c
```diff
@@ -52,7 +53,9 @@ void outmidi_byte(int port, int value) {
 }
 
 // The rest is not relevant to libpd.
```
nettoyeurny: I suppose this line needs to move down now that sys_get_midi_apis does something.

nettoyeurny:
Looks good. One minor nit, otherwise I'm okay with merging this.

notlion:
Heya. I'm curious what's required to get the open patch to display after calling `libpd_start_gui()`. It looks like `sys_pollgui()` needs to be called, but that function doesn't seem to be exposed in the `libpd-osx.a` static library produced by `libpd.xcodeproj`. The GUI sample works for me, but that links against the .dylib produced by the libpd Makefile. I'd be happy to make any modifications to the Xcode project if you could point me in the right direction! Perhaps a `libpd_poll_gui()` function should be added?

millerpuckette:
You're working with my clone of libpd, correct?  (I think there were multiple
problems with anything dating more than a month ago).

I believe everything needed is now exposed in pdlib (and it's wise to go through
the calls in zpdlib.c/h because there might be locks to set).

cheers
Miller

notlion:
@millerpuckette Thanks for the reply. I'm working against the `gui_control` branch from this PR. I tested against your `master` branch though, just to make sure. GUI behavior is the same. After calling `libpd_start_gui()` Wish opens, but neither the Pd main window or patch window are displayed. `sys_pollgui()` is also an undefined symbol in the static library. Maybe it's just the Xcode build that's problematic?

millerpuckette:
Hmm.. quite possibly - I don't know anything about xcode projects or how to
chase down problems in them...  but it's odd that just one external function
would come out undefined.  

I'm finding sys_pollgui gets called in z_libpd.c (libpd code should call
libpd_pollgui() which is a wrapper for it) - and it's defined in the Pd
source.  the client code should periodically call libpd_pollgui() and if
not, the GUI won't appear.

cheers
Miller

notlion:
Aha! I see that `libpd_pollgui()` is implemented in @millerpuckette's `threadable` branch, but not in this PR. After adding that wrapper to my fork everything seems to work!

My branch, which is a fork of `gui_control` for reference:
https://github.com/notlion/libpd/commit/742af024483d81b45130a7f082a2ccc4c4fba25a

richardeakin:
Just wanted to say that I'd love to see this feature make it into master somehow. So much nicer than having to open up your patches in a separate program when debugging.

danomatika:
We will integrate this work, don't worry. Peter and I want to revisit possible API changes for a libpd 1.0 and I'm in a very busy time right now with moving and a kid coming.

danomatika:
This working and was merged with Millers #171 PR. Please test with the current master and open an issue, etc if you run into problems.

Issue #130, opened 2016-04-12
CSharp on Linux: out buffer always zero


residuum:
While trying to figure out the problem with https://github.com/libpd/libpd/pull/113#issuecomment-204458784 I think I have located the bug:

In both [ `libpd_process_raw()`](https://github.com/libpd/libpd/blob/master/libpd_wrapper/z_libpd.c#L138) and the macro [`PROCESS(_x, _y)`](https://github.com/libpd/libpd/blob/master/libpd_wrapper/z_libpd.c#L158) the variable `sys_soundout` is constantly 0.

I have tested by adding a variable count to the functions and a line `if (*p != 0) count++;` in the copy loop (e.g. line 152), returning this variable and then outputting the value to the command line. These functions still return 0.

Subscribing to control rate receivers is no problem.

Here are my changes to the code for the test:

    int libpd_process_raw(const float *inBuffer, float *outBuffer) {
      size_t n_in = sys_inchannels * DEFDACBLKSIZE;
      size_t n_out = sys_outchannels * DEFDACBLKSIZE;
      t_sample *p;
      size_t i;
      size_t count = 0;
      sys_microsleep(0);
      for (p = sys_soundin, i = 0; i < n_in; i++) {
        *p++ = *inBuffer++;
      }
      memset(sys_soundout, 0, n_out * sizeof(t_sample));
      SCHED_TICK(sys_time + sys_time_per_dsp_tick);
      for (p = sys_soundout, i = 0; i < n_out; i++) {
        if (*p1 != 0) count++; \
        *outBuffer++ = *p++;
      }
      return count;
    }

and

    #define PROCESS(_x, _y) \
      int i, j, k; \
      int count = 0;\
      t_sample *p0, *p1; \
      sys_microsleep(0); \
      for (i = 0; i < ticks; i++) { \
        for (j = 0, p0 = sys_soundin; j < DEFDACBLKSIZE; j++, p0++) { \
         for (k = 0, p1 = p0; k < sys_inchannels; k++, p1 += DEFDACBLKSIZE) {\
            *p1 = *inBuffer++ _x; \
          } \
        } \
        memset(sys_soundout, 0, sys_outchannels*DEFDACBLKSIZE*sizeof(t_sample)); \
        SCHED_TICK(sys_time + sys_time_per_dsp_tick); \
        for (j = 0, p0 = sys_soundout; j < DEFDACBLKSIZE; j++, p0++) { \
[...]

PR #139, opened 2016-06-16
Set number format to "C" on pd_init().

residuum:
Fixes issue #130

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,11 @@ ifeq ($(DEBUG), true)
 	OPT_CFLAGS = -Wall
 endif
 
+LOCALE_CFLAGS =
+ifeq ($(LIBPD_SETLOCALE), true)
+	LOCALE_CFLAGS = -DLIBPD_SETLOCALE
+endif
+
 # object files which are somehow generated but not from sources listed above,
 # there is probably a better fix but this works for now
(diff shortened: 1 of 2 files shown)


danomatika:
Maybe this should be a define, in case you need to turn if off? I think it should be on by default. Perhaps something like LIBPD_NO_LOCALE? pinging @nettoyeurny

nettoyeurny:
I agree that there should be a way to turn this off. Not sure whether it should be on by default --- do we know how it interacts with iOS or Android?

residuum:
OK, I have switched to using a define.

danomatika:
This looks good.

One thing I can think of is making the define name less generic. As it applies only to libpd itself, how about `LIBPD_SETLOCALE` or `LIBPD_SET_LOCALE`? This would fit in with the existing `LIBPD_EXTRA`.

danomatika:
This is now merged and I added info to the readme. Make variable is "SETLOCALE" and define is "LIBPD_LOCALE".

danomatika:
Note: Based on #180, I decided to change the default to set LC_NUMERIC. There is info in the readme about this and the compiler define is now "LIBPD_NO_LOCALE" as it is now used to negate this default.

PR #152, opened 2016-10-23
[C#] Update NuGet package to latest version.

residuum:
I have updated the NuGet package with the latest release, and the version change should be reflected in the main repo as well.


danomatika:
I'll merge this, but we'll need to update this info again for the upcoming release.

Issue #153, opened 2016-10-24
Restructuring the C# API


residuum:
The C# API in its current form is not idiomatic: It is just an abstraction of the C API into one large static class. It does the job, but sticks out like a sore thumb, when you try to use it.

I have built some small demo and test projects, and have also started to [document](https://github.com/libpd/libpd/wiki/C%23) the API.

I would like to keep the current API alive, as [libpd4unity](https://github.com/patricksebastien/libpd4unity) et al. depend on it. The new API would be a supplement to the existing one.

I would like to change that, and propose the following structure:
- The new API should return objects for loaded patches and bound receivers. 
- These new classes should be `IDisposable` to facilitate releasing of resources (`Patch.Dispose()` should call `libpd_closefile()`, etc.).
- PdManager in the Python API is an example of a new class that I have in mind.
- Static methods should be reduced to a minimum, maybe even eliminated.

Is anyone willing to help with this project, even only for a discussion of the new design?

danomatika:
Most of the language wrapper find a balance between sticking with the C API and for adding language-specific support, etc. I think your proposal sounds like such.

The C++ wrapper, for instance, wraps the C API but through classes and provides convenience classes around files and lists. It also avoids a global static class in order to hopefully support multi instances later on.

Also, APIs change over time so I think it's reasonable to make improvements and mark things deprecated for a period of time, then remove the old interfaces.

nettoyeurny:
Quick note on timing: I entirely agree that this is a worthwhile project, but the real question is when to do this. While the current C API of libpd has been holding up really well, it does reflect Pd as it was six years ago. I've been waiting for two substantial changes to Pd, multi-instance support and a clear separation between the GUI and the audio core.

Once those are in place (multi-instance support seems to be well on its way), that'll be an opportunity to create libpd 2.0, and then it'll make a lot of sense to get rid of all the static stuff in the object-oriented bindings.

(I'm not objecting to doing this now, just pointing out that if we do this now, we may end up doing it yet again in the not too distant future.)

residuum:
@nettoyeurny I would say that restructuring of the C# API should be done now, while deprecating the old API, so that all users already use the non-static, multi-instance ready version.

danomatika:
The same is true of the Obj-C wrapper which uses a main static class. I'm fine with waiting for an overall API update of that as people with legacy stuff can stick with "libpd 0.#.#" and then later port to "libpd 1.#.#".

danomatika:
Close via merging #154.

PR #154, opened 2016-10-27
Restructuring of C# API

residuum:
As proposed in https://github.com/libpd/libpd/issues/153 I have started restructuing of the C# API.

The work is complete.


csharp/Managed/Patch.cs
```diff
@@ -0,0 +1,44 @@
+using LibPDBinding.Native;
+using System;
+
+namespace LibPDBinding.Managed
+{
+	/// <summary>
+	/// Pd Patch.
+	/// </summary>
+	public class Patch : IDisposable
```
residuum: Patch is IDisposable, so that it is clear, that resources have to be released

csharp/Managed/Pd.cs
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		/// <summary>
+		/// Initializes a new instance of Pd
+		/// </summary>
+		/// <param name="inputChannels">Number of input channels.</param>
+		/// <param name="outputChannels">Number of output channels.</param>
+		/// <param name="sampleRate">Sample rate for project.</param>
+		/// <param name="searchPaths">Paths for Pd to search for externals.</param>
+		public Pd (int inputChannels, int outputChannels, int sampleRate, IEnumerable<string> searchPaths)
```
residuum: OK, so this is a design decision:
- should the consumer of this class be able to change the number of input and output channels later?
- should she be able to add search paths for externals later?

csharp/Managed/Pd.cs
```diff
@@ -0,0 +1,202 @@
...
+			return outBuffer;
+		}
+
+		/// <summary>
+		/// Reads samples from inBuffer and processes them with Pd.
+		/// </summary>
+		/// <param name="ticks">Number of ticks to process. To reduce overhead of function calls, raise this number. To lower latency, reduce this number.</param>
+		/// <param name="inBuffer">Interleaved input buffer. Must be at least of size [InputChannels] * [Blocksize] * [ticks]</param>
+		/// <returns>Interleaved processed audio samples.</returns>
+		public float[] Process (int ticks, float[] inBuffer)
```
residuum: - The `libpd_process_*` functions do never return anything different than 0 (see in C code).
- All pointer versions have been removed. Has anyone ever used them?
- Returning a new array looks more natural to me
- Should ticks or inBuffer be the first parameter?

csharp/Managed/Pd.cs
```diff
@@ -0,0 +1,202 @@
...
+			IsComputing = false;
+		}
+
+		/// <summary>
+		/// Loads a Pd patch from the specified file path.
+		/// </summary>
+		/// <param name="path">Path to the Pd file.</param>
+		/// <returns>New Patch, if loading is successful, else null.</returns>
+		public Patch LoadPatch (string path)
+		{
```
residuum: Compare the comments to class `Patch`.

csharp/Managed/Pd.cs
```diff
@@ -0,0 +1,202 @@
...
+			if (!File.Exists (path)) {
+				return null;
```
residuum: Could throw a `FileNotFoundException` as well.

csharp/Managed/Pd.cs
danomatika: In the other wrappers, we generally try to stick with the argument order from the C api.

csharp/Managed/Pd.cs
```diff
@@ -0,0 +1,202 @@
...
+			SendMessage ("pd", "dsp", 1);
+			IsComputing = true;
+		}
+
+		private void SendMessage (string receiver, string message, params object[] args)
+		{
+			SendArgs (args);
+			var finish = PInvoke.finish_message (receiver, message);
+
+			if (finish != 0) {
```
danomatika: I'd say stick with return codes, they may be uncommon but they match what comes from the C api. Introducing exceptions is probably overkill, at least from my C++ experience.

residuum: The reason behind not using error codes: https://blogs.msdn.microsoft.com/kcwalina/2005/03/16/design-guidelines-update-exception-throwing/

> Do not return error codes. Exceptions are the primary means of reporting errors in frameworks.
> Do report execution failures by throwing exceptions. If a member cannot successfully do what is designed to do, it should be considered an execution failure and an exception should be thrown.

csharp/Managed/Pd.cs
danomatika: Probably should be able to losing things, although I am not familiar with the C# consumer idiom.

danomatika: *modify

csharp/Managed/Patch.cs
```diff
@@ -0,0 +1,44 @@
...
+		public void Dispose ()
+		{
+			Dispose (true);
+			GC.SuppressFinalize (this);
+		}
+
+		private void Dispose (bool disposing)
+		{
+			if (disposing) {
+				PInvoke.closefile (_handle);
```
residuum: Sorry, this is wrong: `_handle` is not a managed resource, so no `if (disposing)` would be correct.

csharp/Managed/Pd.cs
danomatika: Wel, you asked, but it seems like you already have an answer.

residuum: I was more thinking of people familiar with C# and its idioms. Maybe I am missing some crucial point here, and different semantics should be applied.

csharp/Managed/PdArray.cs
```diff
@@ -0,0 +1,43 @@
+﻿using LibPDBinding.Native;
+
+namespace LibPDBinding.Managed
+{
+	public class PdArray
```
residuum: Any idea for a better name? Maybe PdArrayAccess?

csharp/Managed/Messaging.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Runtime.CompilerServices;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace LibPDBinding
+{
+	/// <summary>
+	/// Messaging in Pd.
+	/// </summary>
+	public sealed class Messaging
```
residuum: To minimize methods on a class, I have created a class simple for messaging, that is created with the Pd instance and cannot be created from a user (internal constructor)

csharp/Managed/Messaging.cs
```diff
@@ -0,0 +1,181 @@
...
+			MessageInvocation.SendMessage (receiver, message, atoms);
+		}
+
+		/// <summary>
+		/// Send a message to the specified receiver with a range of atoms.
+		/// </summary>
+		/// <param name="receiver">Receiver.</param>
+		/// <param name="atoms">Atoms.</param>
+		[MethodImpl (MethodImplOptions.Synchronized)]
+		public void Send (string receiver, params IAtom[] atoms){
```
residuum: A single method for sending atoms and list of atoms to Pd.

csharp/Managed/Messaging.cs
```diff
@@ -0,0 +1,181 @@
...
+			MessageInvocation.SendList (receiver, atoms);
+		}		
+
+		/// <summary>
+		/// Send a bang message to the specified receiver with a range of atoms.
+		/// </summary>
+		/// <param name="receiver">Receiver.</param>
+		/// <param name="atoms">Atoms.</param>
+		[MethodImpl (MethodImplOptions.Synchronized)]
+		public void Send (string receiver, Bang bang){
```
residuum: And an overload for sending a Bang message.

csharp/Managed/Events/FloatEventArgs.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using LibPDBinding.Managed.Data;
+
+namespace LibPDBinding.Managed.Events
+{
+	public class FloatEventArgs : EventArgs
+	{
+		public string Receiver { get; private set; }
+
+		public Float Float { get; private set; }
```
residuum: This could also contain a C# `float` instead of the `IAtom` encapsulation, see comment to ListEventArgs.cs.

csharp/Managed/Events/SymbolEventArgs.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using LibPDBinding.Managed.Data;
+
+namespace LibPDBinding.Managed.Events
+{
+	public class SymbolEventArgs : EventArgs
+	{
+		public string Receiver { get; private set; }
+
+		public Symbol Symbol { get; private set; }
```
residuum: This could also contain a C# `string` instead of the `IAtom` encapsulation, see comment to ListEventArgs.cs

csharp/Managed/Events/ListEventArgs.cs
```diff
@@ -0,0 +1,21 @@
...
+using LibPDBinding.Managed.Data;
+using LibPDBinding.Managed.Utils;
+
+namespace LibPDBinding.Managed.Events
+{
+	public class ListEventArgs : EventArgs
+	{
+		public string Receiver { get; private set; }
+
+		public IEnumerable<IAtom> List { get; private set; }
```
residuum: Alternative could be `IAtom[]` or `IEnumerable<object>` or `object[]`. 

- 'IEnumerable` instead of `[]`: most C# API use it and is more like Pd lists look like from coding
- `IAtom` instead of `object`: makes it clear what kind of items can be in the list.

csharp/Managed/Events/MessageEventArgs.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace LibPDBinding.Managed.Events
+{
+	public class MessageEventArgs : EventArgs
+	{
+		public string Receiver { get; private set; }
+
+		public string Message { get; private set; }
+
+		public IEnumerable<IAtom> List { get; private set; }
```
residuum: See ListEventArgs.

csharp/Managed/Events/PrintEventArgs.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using LibPDBinding.Managed.Data;
+
+namespace LibPDBinding.Managed.Events
+{
+	public class PrintEventArgs : EventArgs
+	{
+		public Symbol Symbol { get; private set; }
```
residuum: See SymbolEventArgs.

samples/csharp/pdtest/Managed/AudioTests.cs
```diff
@@ -0,0 +1,112 @@
...
+		{
+			int blocksize = _pd.BlockSize;
+			int ticks = 2;
+			short[] inBuffer = new short[ticks * _inputs * blocksize];
+			for (int i = 0; i < inBuffer.Length; i++) {
+				inBuffer [i] = (short)i;
+			}
+			_pd.Start ();
+			short[] outBuffer = _pd.Process (ticks, inBuffer);
+			for (int i = 0; i < outBuffer.Length / 3; i++) {
```
residuum: The assertions fail, but the test is flawed (that's what they all say, don't they?)

samples/csharp/pdtest/Managed/AudioTests.cs
```diff
@@ -0,0 +1,112 @@
...
+				Assert.AreEqual (2 * i, outBuffer [3 * i], 0.0001);
+				Assert.AreEqual (-6 * i, outBuffer [3 * i + 1], 0.0001);
+				Assert.AreEqual (Math.Cos (2 * Math.PI * 440 / 44100 * i), outBuffer [3 * i + 2], 0.0001);
```
residuum: The assertions fail, but the test is flawed (that's what they all say, don't they?)

samples/csharp/pdtest/Managed/CApiTests.cs
```diff
@@ -0,0 +1,42 @@
...
+		public virtual void TestLoadBang ()
+		{
+			_pd.Messaging.Bind ("foo");
+			string value = "";
+			_pd.Messaging.Print += delegate(object sender, PrintEventArgs e) {
+				value += e.Symbol.Value;
+			};
+			using (Patch patch = _pd.LoadPatch (@"../../test_csharp.pd")) {
+				;
+			}
```
residuum: We only need to load the patch and then dispose of it to get the loadbang.

residuum:
I have just rebased everything onto master and unit tests are still green.

danomatika:
>I have just rebased everything onto master and unit tests are still green.

So this is good to merge?

residuum:
In the current state this should not be merged, it needs a lot of change to the documentation, and not every function has been mapped to the new C# API.

danomatika:
Ok. I'm leaning toward releasing soon rather than later, so we could push this back to the next version.

residuum:
The non-multiinstance API has been migrated to object oriented classes instead of the old static API.

Some questions remain:
1. Should I squash those commits before you merge the pull request?
2. Can someone review the new API? I am thinking of @tebjan @ATHellboy @anthonyburchell 
3. Should we mark the static API as deprecated in code and recommend the object oriented classes to developers?

csharp/Managed/Pd.cs
```diff
@@ -0,0 +1,183 @@
...
+			get { return PInvoke.blocksize (); }
+		}
+
+		Messaging _messaging;
+
+	    /// <summary>
+		/// Gets the messaging object.
+		/// </summary>
+		/// <value>The messaging object.</value>
+		public Messaging Messaging {
```
residuum: Encapsulation of sending messages to and from Pd.

csharp/Managed/Pd.cs
```diff
@@ -0,0 +1,183 @@
...
+				return _messaging;
+			}
+		}
+
+	    Midi _midi;
+
+        /// <summary>
+        /// Get the object for MIDI communication
+        /// </summary>
+	    public Midi Midi
```
residuum: Encapsulation of sending MIDI to and from Pd.

danomatika:
Hw is this coming? I'm thinking of making a new release relatively soon.

residuum:
From my point of view this is ready, all unit tests pass on Windows and Linux. As nobody is commenting on the class structure, I think it can be merged. I will then update the wiki page https://github.com/libpd/libpd/wiki/C%23

danomatika:
>Should I squash these commits into a single one?

Nah. Not so many.

PR #157, opened 2016-11-12
Add PortAudio support

gohai:
This is Peter's portaudio branch squashed to apply cleanly against master. Tested on ARM Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,12 +7,13 @@ ifeq ($(UNAME), Darwin)  # Mac
   PDNATIVE_PLATFORM = mac
   PDNATIVE_ARCH = 
   PLATFORM_CFLAGS = -DHAVE_LIBDL -arch x86_64 -arch i386 -g \
-    -I/System/Library/Frameworks/JavaVM.framework/Headers
+    -I/System/Library/Frameworks/JavaVM.framework/Headers -I/usr/local/include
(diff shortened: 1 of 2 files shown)


nettoyeurny:
Not sure it's a good idea to unconditionally switch the Java version to PortAudio. I would prefer to make this configurable so that users can choose whether they want to use PortAudio.

gohai:
@nettoyeurny That makes a lot of sense to me. Just thought point out this branch that makes it easier to compile against current master with your PortAudio work.

danomatika:
I agree in that we do not want to change the default. I went ahead and edited this PR into a new branch: java_pa

This makes building the portaudio for the jni selectable via the "PORTAUDIO" makefile option. Also, I removed the include path for `/usr/local/lib` as that can be added manually through make via CFLAGS:

    make CFLAGS="-I/usr/local/lib"

There's no guarantee users on macOS will have installed pa from Homebrew, maybe they use MacPorts which installs to /opt. In this case we'd either have to add *all* possible paths or simply leave it up to the user to add manually. I think it makes sense to do the latter for now.

nettoyeurny:
Thanks, Dan! Quick reality check: It looks like you're always adding -lportaudio to the linker flags, even when the PortAudio flag is false. Is this intended?

danomatika:
No, I just forgot to remove them. Fixed now.

PR #160, opened 2016-11-22
Java + PortAudio

nettoyeurny:
I think this is okay to merge. What say you, Dan?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,7 @@ ifeq ($(UNAME), Darwin)  # Mac
   CSHARP_LDFLAGS = $(LDFLAGS)
   CPP_FLAGS = -stdlib=libc++
   CPP_LDFLAGS = $(LDFLAGS) -stdlib=libc++
-  JAVA_LDFLAGS = -lportaudio -framework CoreAudio -framework AudioToolbox -framework AudioUnit \
-    -framework CoreServices -framework JavaVM $(LDFLAGS)
+  JAVA_LDFLAGS = -framework JavaVM $(LDFLAGS)
 else
   ifeq ($(OS), Windows_NT)  # Windows, use Mingw
     CC = gcc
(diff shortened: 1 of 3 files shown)


danomatika:
Can we leave off the explicit `/usr/local/lib` include? Or better yet, add it if the folder exists? This is what I do in the libpd configure script.

nettoyeurny:
Done. I just deleted it; right now I just want to get the PortAudio code into the main branch.

Makefile
```diff
@@ -146,8 +159,9 @@ PDJAVA_DIR = $(PDJAVA_BUILD)/org/puredata/core/natives/$(PDNATIVE_PLATFORM)/$(PD
 PDJAVA_NATIVE = $(PDJAVA_DIR)/$(SOLIB_PREFIX)pdnative.$(PDNATIVE_SOLIB_EXT)
 PDJAVA_JAR = libs/libpd.jar
 
-CFLAGS = -DPD -DHAVE_UNISTD_H -DUSEAPI_DUMMY -I./pure-data/src -I./libpd_wrapper \
-         -I./libpd_wrapper/util $(PLATFORM_CFLAGS) $(OPT_CFLAGS) $(EXTRA_CFLAGS) $(LOCALE_CFLAGS)
+CFLAGS := -DPD -DHAVE_UNISTD_H -DUSEAPI_DUMMY -I./pure-data/src -I./libpd_wrapper \
+         -I./libpd_wrapper/util $(PLATFORM_CFLAGS) $(OPT_CFLAGS) $(EXTRA_CFLAGS) $(LOCALE_CFLAGS) \
+         $(CFLAGS)
```
danomatika: I append $CFLAGS here, does the ":=" do the same?

nettoyeurny: The difference is that := expands all the variables right here (which is okay because none of them will change after this assignment; I checked). The = operator expands later, creating a dependency cycle.

danomatika:
Ok. Looks good.

nettoyeurny:
Actually, I noticed another cyclic dependency. Just a sec...

danomatika:
I already merged, If you're just doing clean up, go ahead and commit to master.

nettoyeurny:
It's more than a cleanup, but I think I'll commit to master anyway because the current state is pretty broken and the significant changes are to z_jni_pa.c, which nothing else depends on.

Issue #159, opened 2016-11-22
Makefile install in sandbox


wimmuskee:
I'm trying to package libpd for Gentoo, and when compiling and installing a package, Gentoo uses a sandbox to do the work before moving the result to the actual filesystem.

To make this work I had to do two things in the Makefile:
- prefix the prefix declaration with $(DESTDIR), to be able to provide Gentoo's sandbox root
- make a $(prefix)/lib dir, in the sandbox it does not exist, and otherwise the library file is copied to /lib, not in /lib

If the Makefile is currently working as intended, close this issue, otherwise I can provide a merge request.

wimmuskee:
never mind on the $(DESTDIR), simply providing the prefix=$(DESTDIR)/usr when calling to install works

danomatika:
The install target is not super fancy right now, for sure. We're thinking about moving the project to Autotools which provides most of this functionality in a more standard way.

PR #161, opened 2016-11-23
makefile: add mkdir libdir, #159

wimmuskee:
#159 

create libdir before copying to it, in case of sandbox installations where the libdir does not exist


danomatika:
Thanks

PR #162, opened 2016-11-26
add working pyaudio python sample

wimmuskee:
I got none of the included pyaudio python samples working. Found the following working one on Stackoverflow, and decided it was a good idea to include it. 

Don't know (yet) why the default samples are not working. I suspect it has something to do with using the callback function.

Until we know, I thought it well to have at least one working pyaudio example listed for other potential users.


danomatika:
Can the default samples be modified in this way to work?

wimmuskee:
probably, but haven't had time to delve into the differences yet (and why to current ones do not work)

but, as soon as I have, I'll put up a merge request for those, closing this one for now

nuromantix:
Did anyone ever figure out why they don't work?

PR #165, opened 2017-03-17
Sets MixWithOthers option for Playback category

alexandernaut:
Ran into a problem where PdAudioController was interrupted/disabled when music started playing from SoundCloud on Safari. I believe this fixes the problem :]

### DIFF
--- a/objc/PdAudioController.m
+++ b/objc/PdAudioController.m
@@ -181,7 +181,19 @@
 		AU_LOG(@"failed to set session category, error %@", error);
 		return PdAudioError;
 	}
-	if ([category isEqualToString:AVAudioSessionCategoryPlayAndRecord]) {
+	
+	
+	if ([category isEqualToString:AVAudioSessionCategoryPlayback]) {
+		if ([[AVAudioSession sharedInstance] respondsToSelector:@selector(setCategory:withOptions:error:)]) {
+			// iOS 7+
+			if (![[AVAudioSession sharedInstance] setCategory:category withOptions:AVAudioSessionCategoryOptionMixWithOthers error:&error]) {
+				AU_LOG(@"error setting AVAudioSessionCategoryOptionMixWithOthers: %@", error.localizedDescription);
+				return PdAudioError;
+			}
+		}
+	}
+	
+	else if ([category isEqualToString:AVAudioSessionCategoryPlayAndRecord]) {
 		if ([[AVAudioSession sharedInstance] respondsToSelector:@selector(setCategory:withOptions:error:)]) {
 			// iOS 7+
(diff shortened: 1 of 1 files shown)


danomatika:
Looks like you could simplify this by changing line 196 to:

    if ([category isEqualToString:AVAudioSessionCategoryPlayback] || [category isEqualToString:AVAudioSessionCategoryPlayAndRecord]) {

The only thing being set in there is to use the speaker by default and to allow mixing, both of which are probably applicable to Playback as well as PlayAndRecord.

alexandernaut:
Hmm, I believe the DefaultToSpeaker option is only applicable to the PlayAndRecord category.  Config fails on my end when setting it on the Playback category.

danomatika:
Ok, thanks for checking. If you can add the iOS 6 part to your PR, we can accept.

alexandernaut:
Just updated, but I'm honestly unsure if I'm using Git correctly (this is my first time). Did you get the changes?

Also, do you have any recommendations for how to best use git in harmony with Xcode? 

In any case, thanks for taking the time to look at this :]

danomatika:
Thanks. Looks cool. I might make a few small changes.

danomatika:
> Also, do you have any recommendations for how to best use git in harmony with Xcode?

I don't, I use git for the commandline. Xcode's diff view is useful though.

danomatika:
Also, how does endInterruptionWithFlags: need to be updated?

alexandernaut:
Woo, thanks for the merge! 🎉🎉

And thanks for all the work you've done with LibPd. I used it for my first app, Arpeggionome Pro, and it made my transition from pd user to app development possible :]

Regarding the update to endInterruptionWithFlags:, the problem is that the AVAudioSessionInterruptionOptionShouldResume flag is deprecated and not set for at least iOS 10.3. So right now, when it gets the call back after an interruption, pd is not turned back on. So it needs to be updated to just use the method callback (and not the flag) for iOS 7 (or iOS6? need to check) and later. 

Also, to filter based on the running OS, I believe the __IPHONE_OS_VERSION_MIN_REQUIRED is not the right thing to use. I'm pretty sure that's just checking the minimum sdk set in Xcode.

alexandernaut:
If you'd like, I can make this change later tonight or tomorrow :]

danomatika:
Sure. Please do. I maybe merged too quickly as I saw the TODO afterwards.

danomatika:
Also, you will need to pull from the main repo into your fork as I made a commit to clean up some whitespace issues.

alexandernaut:
Great, will do. And I think it's a good thing you merged already. Better to have it in the state it's in right now (where Pd does not get interrupted at all) than it getting interrupted and not ever turning back on.

PR #178, opened 2017-07-21
Fix crash libpd_openfile due to lock

pierreguillot:
It seems that `libpd_openfile()` needs to lock the global pd's mutex for writing otherwise because `glob_evalfile()` uses static variable and if several instances call this function it crashes.


danomatika:
Thanks for checking this. Do we need *both* sys_lock and the globallock? I don't know, just checking.

pierreguillot:
Yes because `sys_lock()` lock the instance and enables `pd_this->pd_islocked = 1;` then `globallock()` checks this value `if (!pd_this->pd_islocked) bug("pd_globallock");`

pierreguillot:
I can also add my [test](https://github.com/pierreguillot/libpd/blob/dev/threadable/samples/c/pdtest_thread/pdtest_thread.c) (samples/c/pdtest_thread) from this [branch](https://github.com/pierreguillot/libpd/tree/dev/threadable) but it's a bit dirty for the moment

danomatika:
Adding a test is a good idea. I recently cleaned up the makefiles so they they should be easy to adapt. The plan in the future is to update libpd to use auto tools.

danomatika:
Looks good. Let me know when you feel the PR is ready. I'm thinking we will also provide a libpd wrapper for `pd_instance` such as `libpd_set_instance` or something similar.

pierreguillot:
I think it's ready for this PR. Sorry for the slowness :)

danomatika:
If you use the current master branch, you could update the test to use the "libpd_*_instance" functions. Also, the main Makefile now as a "MULTI" build option so you no longer need to set PDINSTANCE and DTHREADS in the test Makefile. Build libpd with multiple instance support via: `make MULTI=true`.

pierreguillot:
Done ! Thanks

PR #182, opened 2017-08-08
Add support for Carthage, and dynamic frameworks

### LINKED ISSUE #163: Carthage support
Would you be interested in [Carthage](https://github.com/Carthage/Carthage) support for Xcode projects?

I have added support in [my `libpd` fork](https://github.com/squarefrog/libpd/tree/carthage) so I can use it, but wondered if it was of any use to the `libpd` community.

The main disadvantage to adding support would mean the minimum deployment targets raise to iOS 8, although 92% of iOS users are using version [9 or above](https://developer.apple.com/support/app-store/).

If you'd rather not merge in support, then no problem, people can still use Carthage by adding my fork to their `Cartfile`:

    github "squarefrog/libpd" "carthage"

I don't believe this would negatively affect users who are adding the project 'the hard way'.

squarefrog:
This PR adds support for [Carthage](https://github.com/Carthage/Carthage).

This does upgrade the Xcode project to 8.1, so consider if that is something you want. I don't personally see a reason to hold back modern Xcode support for older versions, but it's your call.

To use the `PureData` dynamic framework, either via Carthage or manually, then your minimum deployment target should be iOS 8.

Closes #163


danomatika:
I've gone ahead and updated the Xcode projects for Xcode 9 with a min deployment of iOS 8.0. I've also started modernization of the Obj-C wrapper with ARC support.

squarefrog:
👍

danomatika:
No need to delete this. I only updated the Xcode projects and some of the code. I forgot you had already done part of this before I started.

squarefrog:
OK I'll reopen, but there will probably be conflicts as Xcode 9 wasn't around when I created the PR.

danomatika:
Yeah, you'll need to overwrite your libpd.xcodeproj changes which those in master, either manually or via a merge and then fix the conflicts via `git checkout --theirs -- <filename>` , etc.

danomatika:
If you want this in, you could resubmit from the current Xcode project. I'd ask if the framework can be named "libpd" instead of "PureData".

danomatika:
I'm going to close this for now as I think the branches got out of sync. Go ahead and submit a new PR with the libpd framework and the objc wrapper updates.

PR #183, opened 2017-08-20
Updates to Android.mk

tkirshboim:
- Atomic is now required by ringbuffer.c
- s_file.c is no longer required for libpd (Android.mk counterpart of https://github.com/libpd/libpd/commit/8f08405c40f87ba93e0bc7dba480fc73a1bfeefe )


tkirshboim:
pinging @joebowbeer

danomatika:
>Can you explain why s_file.c is not built? It handles preferences? I see a comment about Android in the source.

It's the desktop preferences which has no real meaning for libpd.

danomatika:
Note that I've introduced prelim function wrappers for multiple instance support. The pd sources need to be build with PDINSTANCE and PDTHREADS defined. This is done with the main makefile "MULTI=true" option.

Multiple instance support may or may not be applicable to Android, but I dunno for sure.

PR #189, opened 2017-09-27
Use the github puredata submodule and only include obj files as public headers

cerupcat:
I can break this into two pull requests if needed, but one fixes the issue where git://git.code.sf.net/p/pure-data/pure-data is offline and therefore you can't pull the submodule.

The other fixes issues with building cocoapod frameworks and using swift since it includes header files that won't compile correctly in the libpd-umbrella.h file. This exposes only the obj folder to the umbrella file fixing build issues.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = git://github.com/nettoyeurny/opensl_stream.git
 [submodule "pure-data"]
 	path = pure-data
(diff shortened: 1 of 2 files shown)


danomatika:
I've added the pod spec update but we prefer to keep the current pure-data upstream remote for now as it's still the main upstream repo. The pure-data github repo is kept in sync most of theme but is still not the main repo *yet*.

PR #198, opened 2017-11-03
Add travis-ci support for C

pierreguillot:
The branch adds *.travis.yml* that makes and runs libpd and the C samples on Linux and MacOS. Here are the [results](https://travis-ci.org/pierreguillot/libpd).


.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+  - cd samples/c/pdtest && make && cd ../../..
+  - cd samples/c/pdtest_gui && make && cd ../../..
+  - cd samples/c/pdtest_multi && make && cd ../../..
+  #- cd samples/c/pdtest_thread && make && cd ../../..
+
+after_success:
+  - cd samples/c/pdtest && ./pdtest test.pd ./ && cd ../../..
+  #- cd samples/c/pdtest_gui && ./pdtest_gui test.pd ./ && cd ../../..
+  - cd samples/c/pdtest_multi && ./pdtest_multi test.pd ./ && cd ../../..
+  #- cd samples/c/pdtest_thread && pdtest_thread ./ test.pd && cd ../../..
```
danomatika: Note: I just pushed a commit which changes pdtest_thread to use the same file & folder arguments as the other C samples. I was having trouble with the paths before.

pierreguillot: Thanks ! I was correcting it  also :) I keep you informed when the branch is ready (there is a problem with pthread on Linux)

pierreguillot:
I think that on macOS, gcc is just a wrapper around clang that doesn't need that pthread to be explicitly linked. I remember I already encountered the same problem. And the linking with pthread on macOS broke my compilation... Here it worked on my computer but you're right it's better to add only to Linux. I also have to check if it's necessary on Windows.

danomatika:
It's necessary on Windows. Check the platform specifics in the main makefile. Worse case is you can link to the libpthread-win in the `libs` folder.

PR #207, opened 2017-11-29
Add raw processing functions for double and short

residuum:
Make the same functionality available as with process for double and short.


libpd_wrapper/z_libpd.h
```diff
@@ -29,6 +29,9 @@ EXTERN int libpd_getdollarzero(void *p);
 EXTERN int libpd_blocksize(void);
 EXTERN int libpd_init_audio(int inChans, int outChans, int sampleRate);
 EXTERN int libpd_process_raw(const float *inBuffer, float *outBuffer);
```
residuum: Keep the existing function for compatibility.

danomatika:
I'm curious, what is this needed for? As in, in what cases does the existing raw function not work?

residuum:
It is not really needed, but currently we have

     EXTERN int libpd_process_short(const int ticks, const short *inBuffer, short *outBuffer);
     EXTERN int libpd_process_float(int ticks, const float *inBuffer, float *outBuffer);
     EXTERN int libpd_process_double(int ticks, const double *inBuffer, double *outBuffer);

and 

     EXTERN int libpd_process_raw(const float *inBuffer, float *outBuffer);

This looks inconsitent to me, so I added 

     EXTERN int libpd_process_raw_short(const short *inBuffer, short *outBuffer);
     EXTERN int libpd_process_raw_float(const float *inBuffer, float *outBuffer);
     EXTERN int libpd_process_raw_double(const double *inBuffer, double *outBuffer);

So that one can also process double and short with libPd without striping.

danomatika:
@nettoyeurny thoughts on this? I don't see a major reason against, although some of the wrappers may need to be extended to cover the new functions, this shouldn't break anything.

danomatika:
@nettoyeurny thoughts?

danomatika:
I'm going to take this while dropping `libpd_process_raw_float`. I agree it looks more consistent to have it, but it's not needed when the original `libpd_process_raw` is already there. Growing pains.

PR #213, opened 2017-12-14
Add support for FreeBSD

lhondareyte:
FreeBSD does not have libdl. Adding specific section for FreeBSD (and Linux) in Makefile. That way,  that we can manage other *NIX systems.
Tested ok with Debian 9.0 and FreeBSD 11.1


danomatika:
Thanks.

Issue #215, opened 2018-01-05
Additional linker flags in the Makefile (for libgcc and pthread on Windows)


pierreguillot:
Using libpd (C version) on Windows requires libgcc.dll and pthread.dll. This dependencies can be avoid if we add the linker flags `-static-libgcc` (for libgcc like in the C# linker flags) and `-static` for pthread. I'm not sure that would be good for all the approach so perhaps a solution would be to define additional linker flags?

danomatika:
It should work via:

    make LDFLAGS="-static-libgcc"

A deeper option would be to add an additional makefile flag ala "STATIC_GCC=true" or something similar.

A longer term option is to transition the project to use autotools so we don't have to do as many of these details ourselves and can provide more standard options via configure.

pierreguillot:
I think that `LDFLAGS="..."`override the default linker flags (on my machine at least). 
So if I don't want to modify the makefile I end up with  `LDFLAGS="$(MINGW_LDFLAGS) -Wl,--output-def=libs/libpd.def -static -static-libgcc -Wl,--out-implib=libs/libpd.lib"`. 

Like for `ADDITIONAL_CFLAGS`, `ADDITIONAL_LDFLAGS`would be good temporary solution (before autotools). The Makefile could be something like:

```
LDFLAGS="$(MINGW_LDFLAGS) $(ADDITIONAL_LDFLAGS) -Wl,--output-def=libs/libpd.def -Wl,--out-implib=libs/libpd.lib"
```
And `-static`is important to link statically pthread.

danomatika:
ADDITIONAL_LDFLAGS would work but I think LDFLAGS should be overridable as they are with configure, ala something like:

    LDFLAGS = $(MINGW_LDFLAGS) -Wl,--output-def=libs/libpd.def \
      -Wl,--out-implib=libs/libpd.lib $(LDFLAGS)

This way the user settings are last.

pierreguillot:
You want me to create a PR?

danomatika:
Sure. if it's working. :)

danomatika:
You also add CFLAGS overrides.

pierreguillot:
Done

PR #216, opened 2018-01-05
add ADDITIONAL_LDFLAGS to Makefile

pierreguillot:
The issue is explained in #215. I tested it on MacOS and Windows (MinGW).


danomatika:
Added via 9fd983507c51d2321d135025d56e989dddc9b189.

Issue #217, opened 2018-01-09
change log level


pierreguillot:
*sys_debuglevel* is set to 0 during libpd initialization (l.75 in z_libpd.c). I think it could be great if we could set this to another value with a funtion like *void libpd_setdebuglevel(int level)* for example. It would help people debugging their patches in Camomile.

danomatika:
Maybe we need a set:

    void libpd_set_loglevel(int level);
    int libpd_get_loglevel();

pierreguillot:
In fact this is sys_verbose that needs to be changeable. I create a PR.

danomatika:
Closed with #218.

PR #218, opened 2018-01-09
Add log level function

pierreguillot:
This PR adds to function in the C wrapper to change the loglevel (sys_verbose) - see #217.

### DIFF
--- a/libpd_wrapper/z_libpd.c
+++ b/libpd_wrapper/z_libpd.c
@@ -72,7 +72,6 @@ int libpd_init(void) {
   sys_externalschedlib = 0;
   sys_printtostderr = 0;
   sys_usestdpath = 0; // don't use pd_extrapath, only sys_searchpath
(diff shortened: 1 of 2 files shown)


danomatika:
A couple changes:

* renamed to libpd_set_verbose()/libpd_get_verbose() to reflect what it actually controls
* keep sys_verbose = 0 in init, call libpd_set_verbose() after calling libpd_init()
* make sure set int is used as a bool

pierreguillot:
- if verbose is init in `libpd_init()`, you can't have the verbose option activated while intialization. I thought that global variable like sys_verbose is init to 0 by default with C? Perhaps there can be another solution?
- sys_verbose is not a bool. Just `sys_verbose  > 0` would be better

danomatika:
Ok fixed. Static ints are indeed inited to 0 in C.

pierreguillot:
Great! Thanks!

PR #220, opened 2018-01-11
Update version for C# dll and NuGet

residuum:
Please merge this before tagging 0.11, but not earlier.


danomatika:
Does this changeling summary make sense? https://github.com/libpd/libpd/blob/master/CHANGES.txt#L35

danomatika:
I updated the README with info on setting up Msys2 & MinGW for building the csharplib on Windows. Can you check it out? I'm able to build both in 32 as well as 64 bit on Windows 10, although I've not tested the dlls.

danomatika:
Also, should we mark the original LibPDBinding as deprecated?

residuum:
> I updated the README with info on setting up Msys2 & MinGW for building the csharplib on Windows. Can you check it out? I'm able to build both in 32 as well as 64 bit on Windows 10, although I've not tested the dlls.

Yes, the library can be loaded and unit tests pass on Windows 8.1 with 32 and 64 bit code.

> Also, should we mark the original LibPDBinding as deprecated?

Good idea. With the update of the wrapper to multiinstance support, this will be better, and better warn people early enough. Should I add another PR or should I push to this one?

danomatika:
You can push here.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Is this ready to merge?

residuum:
This is ready to merge, yes.

danomatika:
Bam

Issue #179, opened 2017-07-21
Cmake & samples inside the projects


pierreguillot:
I have setup Cmake for libpd, it generates the static and dynamic libraries and I can also integrate the examples from the samples folder (at least the c samples). I can create a PR, are you interested?

danomatika:
We're more likely to move to auto tools to avoid a make requirement. Closing for now.

PR #226, opened 2018-02-09
Add CMake build script for MSVC, Linux and macOS

eldruin:
This script can be used to generate a Microsoft Visual Studio solution and a makefile on Linux to build libpd C library and the C example programs.
It could be further extended to support MinGW on Windows (should be trivial), the java bindings, etc.
On Windows the pthread library should be placed appropriately in the build directory.


pierreguillot:
Hi, I already suggested the CMake approach but they would prefer to use autotools (see #179). (and [here](https://github.com/pierreguillot/libpd/tree/dev/static-lib-windows) is the branch with my CMakeLists if you want.

danomatika:
I would be willing to have CMake for generating VS solutions but we really don't need it for other platforms. It could simply be used only for Windows as far as I'm concerned and would need to be support aka *do not* give us something then disappear since we (mainly I at this point) have to try and maintain all of these various platform/build system/dpendency manager files. It's becoming a PITA.

I'm not a fan of having a hard dependency on a tool like Cmake when currently all you need is make & a compiler. Even moving to autotools allows us to generate a dot which only relies on make + build chain.

CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
...
+  list(APPEND PD_SOURCES ${PD_UTILS_FILES})
+endif()
+if(pd_extra)
+  list(APPEND PD_SOURCES ${PD_EXTRA_FILES})
+endif()
+
+add_library(libpd SHARED ${PD_SOURCES})
+if(MSVC)
+  target_link_libraries(libpd PUBLIC Ws2_32.lib)
+  target_link_libraries(libpd PUBLIC pthreadVC2.lib)
```
danomatika: We don't distribute a pthread .lib. Where would this be coming from?

eldruin: In this initial version I just added it locally. The location should probably be a build parameter. Something like PTHREAD_LIB_PATH or so.

danomatika: We distribute a copy of the pthread .dll, so could conceivably include the .lib as well. Newer versions of the win pthread implementation are named winpthread-1 I believe.

pierreguillot: I think, you should use **CMAKE_THREAD_LIBS_INIT** and **CMAKE_THREAD_PREFER_PTHREAD**, see [this](https://cmake.org/cmake/help/v3.0/module/FindThreads.html)

eldruin:
@danomatika I see, CMake for MSVC on Windows and makefiles for everything else would already be fine for me as I can do something different for each OS. I will use the cmake script internally at least for Windows anyway, I just wanted to contribute it and see if there is interest.
Regarding maintainability, thanks to @pierreguillot I was able to get past #226 so maybe we actually decide to use libpd, although I am still seeing some crashes in the test executables. If we do, I will take care of making libpd compile on every VisualStudio compiler starting with VisualStudio 2015 (VC14) in our organization and would have no problem in contributing any necessary fixes here.
@pierreguillot Thank you for the link. Your CMake script looks good. Maybe I can extend it with mine for MSVC support and DLL generation.

As I said, I am not sure yet if I will integrate libpd but a responsive and helpful community definitely helps already. Maybe you can decide on whether to accept a CMake script (probably an improved version) depending on that decision.

pierreguillot:
I don't really see the advantage of using CMake only for MSVC. It would be simpler to include and to maintain a MSVC project in this case. And if you add CMake, you can be sure that you'll receive issues and PR about using it with Linux, MacOS, etc. (Note that I'm nevertheless pro CMake :) )

danomatika:
>I don't really see the advantage of using CMake only for MSVC.

And I don't see the advantage of CocoaPods only for iOS but hey, we're stuck with it for now.

Let's just say I am skeptical because there is only so much time available for this project and the more build systems we support, the more time is spent maintaining them. The project started with simple makefiles which are now not so simple but working. I'm able to keep the Xcode project working as I regular use Xcode. But at this point, we'll really need more Windows devs to maintain a VS project otherwise I really don't want something put in a PR and then abandoned later on.

It's not a technical problem but a management problem.

In any case, all of this is moot as y'all need to work with Miller upstream to make the pd sources compatible with VS to begin with.

eldruin:
For the record, the CMakeLists.txt works fine on macOS as well.
@pierreguillot I merged it with your version so it should support your use case as well. Would you be interested in a PR to your branch/extending it?

CMakeLists.txt
```diff
@@ -0,0 +1,234 @@
+cmake_minimum_required (VERSION 2.8.11)
+project (libpd C)
+
+option(BUILD_SHARED_LIBS "Build libpd as a shared library instead of static" ON)
```
pierreguillot: Why not both directly?

danomatika:
Ok, sweet. If y'all can do the testing and confirm things are working, I've decided I'll add the cmake file as an alternative build system instead of replacing the current makefiles. It's not a bad idea to support different workflows, we just need help in maintaining them.

Can you add some basic info to the readme which explains how to use it to, ie. generate and build a Visual Studio libpd project?

CMakeLists.txt
```diff
@@ -0,0 +1,234 @@
...
+option(pd_utils "Compile utilities" ON)
+option(pd_extra "Compile extras" ON)
+option(pd_multi "Compile with multiple instance support" ON)
```
pierreguillot: I think it should be off by default

CMakeLists.txt
```diff
@@ -0,0 +1,234 @@
...
+        set(CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DPD_LONGINTTYPE=\"long long\"")
+    endif()
+else()
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-int-to-pointer-cast -Wno-pointer-to-int-cast")
+    set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS} -ffast-math -funroll-loops -fomit-frame-pointer -O3")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS} -g -O0")
+endif()
+add_definitions(-DPD=1 -DUSEAPI_DUMMY=1)
+
+if(COVERALLS)
```
pierreguillot: I guess it's not necessary for the original libpd repository

CMakeLists.txt
```diff
@@ -0,0 +1,234 @@
...
+
+source_group(pd FILES ${PDSOURCES})
+source_group(pdextra FILES ${PDEXTRASOURCES})
+source_group(libpd FILES ${LIBPDSOURCES})
+source_group(libpdutils FILES ${LIBPDUTILSSOURCES})
+
+include_directories(${PROJECT_SOURCE_DIR}/libpd_wrapper)
+include_directories(${PROJECT_SOURCE_DIR}/pure-data/src)
+
+if(WIN32)
```
pierreguillot: This path is not generic (it was where I compiled the lib in my old project)

CMakeLists.txt
pierreguillot: By the way, the static has been totally removed from my CMakeList

CMakeLists.txt
danomatika: Yes. It should be an option but not default behavior.

CMakeLists.txt
danomatika: I agree. Building both makes sense for Windows targets.

pierreguillot:
@eldruin I'm afraid I can't really try it like this as you removed some essential features like the static lib. Anyway, when I proposed to use CMake as they refused I didn't work on a proper/common approach. So the CMakeList files on my projects have been created specifically for these projects and can't be used directly for a general approach. I don't think it should be used directly. Perhaps we could define how the CMake approach should be done and what it should offer and then try to find the better way?

danomatika:
What I'm proposing is to accept the CMakeLists.txt as an alternative build system. As far as I can tell, there isn't really anything wrong with supporting both, other than some additional maintenance. The overall C lib is pretty stable, so it will not be a major issue.

If y'all can work out a way to where the current Makefiles are not overwritten, it should be fine. In the end, we all just want to be able to build and use libpd to suit our needs, so I'm willing to support that in the end.

pierreguillot:
And I'm just telling that my personal CMakeLists.txt are a bit dirty and to avoid too many additional maintenance we should perhaps create some kind of feature list that defines what people would expect in general from the CMake approach

CMakeLists.txt
eldruin: BUILD_SHARED_LIBS is a standard toggle flag in cmake. if you set it to OFF, it will build a static library.
We can also build both at the same time, but static and dynamic libraries share a .lib file, which means you need to have a different file name.
Normally this is solved by generating something like pd.dll, pd.lib (for dynamic libs) and pd-static.lib
Would you like that?

CMakeLists.txt
eldruin: sure

CMakeLists.txt
eldruin: Sure. I will remove it. I tried to keep the functionality of your script as well.

CMakeLists.txt
eldruin: Sure, I can remove it. I only kept it to keep it working for you.
What I do is to provide CMAKE_THREAD_LIBS_INIT and PTHREADS_INCLUDE_DIR as parameters to cmake, which allows me to link to a custom-built pthreads dll. This is the standard way to do it with the cmake pthreads module. I can document this on a "how to build with cmake".

eldruin:
Ok this sounds great. I will cleanup/adapt the cmake script soon and we can refine it again, then.

CMakeLists.txt
pierreguillot: I think you should also use `set(CMAKE_THREAD_PREFER_PTHREAD True)`

CMakeLists.txt
pierreguillot: Perhaps more libpd.dll/libpd.lib & libpd-static.lib?  But libpd-static should be only for Windows, Linux and MacOS use different extensions for dynamic and static libraries. Another solution would be to use different output folders.

pierreguillot:
Suggestions/Questions
- I think it's create to know where the libs are installed whatever the build system used by CMake. By default libpd use the libs folder I think. Can we keep it from my files?
- Perhaps multi instance support should change the lib names to avoid conflict with the other libpd library?
- In my files, I use `${PROJECT_SOURCE_DIR}` before each path but I'm not sure that's the best thing to do. Perhaps relative paths are more elegant/flexible?

CMakeLists.txt
danomatika: "libpd.dll" as the Pure Data desktop app builds a "pd.dll" Also "libpd-static.dll/lib" sounds good to me.

CMakeLists.txt
eldruin: Yes, I also provide that flag to cmake

eldruin:
@pierreguillot 
# The first point I do not understand
# I could change the library name to libpd_multi.dll/lib and libpd_multi_static.lib/a or something similar if multiinstance is activated. Which name do you want exactly?
# I use relative paths now.

pierreguillot:
Sorry, I mean the output location will be different if you use an XCode project or if you use a Makefile. In my CMakeList.txt I used to "force" the output location to the libs folder. But, like the rest, it's only a suggestion and can be discussed :) Anyway thanks for your work!

danomatika:
I think CMake-generated makefiles should place the built libs in the libs folder. A Visual Studio project, however, should probably stick to whatever is normal for VS. I also think it would be a good idea for the VS project to be solefcotained, so it could be generated into a subfolder, maybe "vs". This way all of the project files aren't polluting the main libpd folder.

eldruin:
The usual way to compile a project with cmake is to create a `build` folder and run `cmake ..` there. That generates all  files inside the `build` folder, including output files which are per default located in `build/<Release|Debug|...>`. e.g. `build/Release/libpd.dll`, `build/Release/pdtest.exe` 
Now both release and debug output files will be put in the `<libpdroot>/libs` folder, including the example C programs, which need to be next to the libpd dlls.
I could also link libpd statically put the example programs in a new `bin` folder.

pierreguillot:
> I could change the library name to libpd_multi.dll/lib and libpd_multi_static.lib/a or something similar if multiinstance is activated. Which name do you want exactly?

It seems good to me! What do you think @danomatika ?

I just tried on Linux, MacOS and Windows and it fails on all platforms. I'll try to see what is different from my file and create a PR to your branch.

danomatika:
Which name do you want exactly?

I would do:

* libpd.dll
* libpd-static.dll
* libpd-multi.dll
* libpd-multi-static.dll

eldruin:
The names are now set as agreed.
@pierreguillot please let me know which errors you are seeing.

pierreguillot:
On Linux and MacOS the names don't need the **-static**.
I don't have access to all OS all the time, on Mac I guess this is mostly on architecture/deployment target problem I think.  For Windows and Linux I need to investigate a bit more to find what's wrong. Do you compile and use the static and dynamic libraries on all the OS? You don't have any problem using it?

CMakeLists.txt
```diff
@@ -0,0 +1,222 @@
...
+        set(CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DPD_LONGINTTYPE=\"long long\"")
+    endif()
+else()
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-int-to-pointer-cast -Wno-pointer-to-int-cast")
+    set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS} -ffast-math -funroll-loops -fomit-frame-pointer -O3")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS} -g -O0")
+endif()
+add_definitions(-DPD=1 -DUSEAPI_DUMMY=1)
+
+if(NOT DEFINED CMAKE_OSX_DEPLOYMENT_TARGET)
```
pierreguillot: There is always a false positive

CMakeLists.txt
```diff
@@ -0,0 +1,222 @@
...
+    set(CMAKE_OSX_DEPLOYMENT_TARGET 10.7)
+endif()
+if(NOT DEFINED CMAKE_OSX_ARCHITECTURES)
```
pierreguillot: idem

CMakeLists.txt
```diff
@@ -0,0 +1,226 @@
...
+endif()
+set(CMAKE_MACOSX_RPATH ON)
+
+foreach (_BUILD_TYPE RELEASE DEBUG)
+    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${_BUILD_TYPE} ${PROJECT_SOURCE_DIR}/libs)
+    set(CMAKE_LIBRARY_OUTPUT_DIRECTORY_${_BUILD_TYPE} ${PROJECT_SOURCE_DIR}/libs)
+    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_${_BUILD_TYPE} ${PROJECT_SOURCE_DIR}/libs)
+endforeach()
+
+set(PD_SOURCES
```
pierreguillot: It would have been better to keep the header files I think

eldruin: Header files in the sources is an [antipattern](http://voices.canonical.com/jussi.pakkanen/2013/03/26/a-list-of-common-cmake-antipatterns/)

CMakeLists.txt
eldruin: Sorry, what do you mean?

CMakeLists.txt
pierreguillot: From your reference:
> The reason people put that in is probably because they think it is required to make CMake rebuild the target in case the header is changed.

No, it's because you can see and select them in your project like XCode. It's not a problem to add the header files in the sources with CMake.

eldruin:
- Regarding the `-static` in the names, yes, it is unnecessary on non-windows platforms. Do you prefer to have it only on windows or to have the same output names everywhere?
- I do not use the static libraries. If you want to use them, you will have to link your program/library additionally with the necessary libraries upon which libpd depends:
  - win32: `Ws2_32`, `pdthread<VC2|...>`
  - linux: `m`, `dl`, `pthread`
  - macOS: `dl`, `pthread`

CMakeLists.txt
eldruin: OK after reading the documentation I think the best would be to remove completely the CMAKE_OSX_DEPLOYMENT_TARGET and CMAKE_OSX_ARCHITECTURES.

CMakeLists.txt
eldruin: I see, I could add them again for your convenience.

pierreguillot:
> Do you prefer to have it only on windows or to have the same output names everywhere?

Yes I think **-static** should be only on Windows except if you have a reason not to do it.

> If you want to use them, you will have to link your program/library additionally with the necessary libraries upon which libpd depends

Yes I know I already compile libpd with my CMake ;). 

I think I found the problem on Linux (and I think with GCC in general). You need to generate position independent code (PIC):

```cmake
if(CMAKE_C_COMPILER_ID STREQUAL GNU)
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
```

danomatika:
Why not use the platform flags we have in the Makefile for macOS and Linux...

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

eldruin:
@danomatika yes, that is how I created the original cmake script version.
@pierreguillot PIC is automatically on for shared libraries and I did not think about further use cases. It is always active now.

eldruin:
Do you want any further changes?

pierreguillot:
> PIC is automatically on for shared libraries and I did not think about further use cases. It is always active now.

Are you sure that PIC should be always active? On MacOS static library doesn't need it I think and I think it slow down the code.

> OK after reading the documentation I think the best would be to remove completely the CMAKE_OSX_DEPLOYMENT_TARGET and CMAKE_OSX_ARCHITECTURES.

Why? Perhaps we could use another user variable to define these? I think it's useful and this ensure that the library can be deployed and used on Mac OS 10.5/6 or 7 (or whatever). 
Idem for the architecture. What would be the problem to say by default the library is universal? 
But perhaps I missed something important in the documentation.

Other suggestions:
- In general, it seems that the option are uppercase (`PD_UTILS` for example). What do you think?
- For the thread library I think we can use cmake 3.1 and `Threads::Threads`
- I still don't understand how do you link against pthread on Windows. Don't you have to use `THREADS_PREFER_PTHREAD_FLAG` or `CMAKE_THREAD_PREFER_PTHREAD` ? Otherwise Cmake use the default win32 thread library. `find_package(Threads REQUIRED)` doesn't mean pthread.
- I think some `C_FLAGS` are in fact `compile_definition`. Or perhaps we can use `C_FLAGS` only?
- Xcode project can be really improved using some specific options. And perhaps the same can be done for MSVC? 
- Perhaps a bit of comments about what a line or a section is supposed to do and why?  

That are only suggestion. Thanks!

Last question: What is Bsymbolic? I don't really understand and as I never needed it...

pierreguillot:
>any numbers? 

No, sorry, I don't have any numbers and I must admit that I'm not a specialist. I read that eliminating PIC improves performances ([PIC-Mach-O](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/dynamic_code.html) or [PIC-Shared-Libraries](https://eli.thegreenplace.net/2011/11/03/position-independent-code-pic-in-shared-libraries/)) but indeed, this could be negligible on x64. As the Makefile doesn't use PIC on MacOS, I just wanted to know why this option would be necessary now with CMake and if there would be any consequence. 
PS: It was a question not a requirement to change it.

umlaeute:
np. i'm also interested in learning something new about `PIC`. btw, there's a [follow-up](https://eli.thegreenplace.net/2011/11/11/position-independent-code-pic-in-shared-libraries-on-x64) to one of your references that explicitly deals with x86_64 (concluding that PIC is slightly slower on x86_64 that non-PIC, but much faster than on x86_32 - suggesting that PIC is indeed a good default value for shared libraries on this arch)

eldruin:
@pierreguillot:
* The user can specify `CMAKE_OSX_DEPLOYMENT_TARGET` and `CMAKE_OSX_ARCHITECTURES` if differing settings are desired (target system different than current). This can be specified by calling cmake with `-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.7`. 
  * If you want to ensure compatibility with a specific macOS version per default, I can re-add them.
* On Windows the user needs to specify where the pthreads library is. I have changed the script to actually require these options:
```
-DCMAKE_THREAD_LIBS_INIT:PATH=<path to library, either VC or GNUC version>
-DPTHREADS_INCLUDE_DIR:PATH=...
```
> I think some C_FLAGS are in fact compile_definition. Or perhaps we can use C_FLAGS only?

Which ones?
> Xcode project can be really improved using some specific options. And perhaps the same can be done for MSVC?

I added some source groups but it already looks fine to me. Improvements are also welcome.
* About -Bsymbolic: I have never needed it. I added it because it is there in the libpd makefile.
* I also added some comments

CMakeLists.txt
```diff
@@ -1,19 +1,36 @@
...
+    option(CMAKE_THREAD_LIBS_INIT "Path to pthreads library")
+    option(PTHREADS_INCLUDE_DIR "Path to the pthreads library header files")
+    # We need pthreads.
+    # Please provide the path to the pthreads library include path and
+    # the path to the pthread library by specifying the following arguments
+    # on the CMake command-line:
+    # -DCMAKE_THREAD_LIBS_INIT:PATH=<path to library, either MSVC (for example pthreadVC2.lib) or GNUC version>
+    # -DPTHREADS_INCLUDE_DIR:PATH=<path to the pthread header files>
+    if (NOT CMAKE_THREAD_LIBS_INIT OR NOT PTHREADS_INCLUDE_DIR)
+        message(FATAL_ERROR "Please provide a path to the pthreads library and its headers.")
```
umlaeute: hmm, assuming you are compiling with MSYS2 on Win32...i think msys2 will properly (provide and) find pthreads on its own, so the test seems to be over-zealous.

how about: checking whether one can use `-lpthread`?

eldruin: I did not find any way to check for -pthread compatibility. Now I only require the pthreads library location when using MSVC.
The user can still provide it if not using MSVC and will be taken into account, in case a different library is desired.

eldruin:
Do you want any further changes?

CMakeLists.txt
```diff
@@ -0,0 +1,260 @@
...
+endif()
+if (PD_MULTI)
+    set(LIBPD_OUTPUT_NAME ${LIBPD_OUTPUT_NAME}-multi)
+endif()
+if (WIN32)
+    set_target_properties(libpd_static PROPERTIES OUTPUT_NAME ${LIBPD_OUTPUT_NAME}-static)
+else()
+    set_target_properties(libpd_static PROPERTIES OUTPUT_NAME ${LIBPD_OUTPUT_NAME})
+endif()
+set_target_properties(libpd        PROPERTIES OUTPUT_NAME ${LIBPD_OUTPUT_NAME})
```
danomatika: Should this line have such a large whitespace gap in the middle?

CMakeLists.txt
```diff
@@ -0,0 +1,260 @@
...
+        if(DEFINED CMAKE_DEBUG_POSTFIX)
+            set_target_properties(${name} PROPERTIES DEBUG_POSTFIX ${CMAKE_DEBUG_POSTFIX})
+        endif()
+    endmacro(ADD_EXAMPLE)
+
+    ADD_EXAMPLE(pdtest)
+    if(PD_MULTI)
+        ADD_EXAMPLE(pdtest_multi)
+        if(NOT MSVC)
+            # uses gettimeofday()
```
danomatika: Instead of bypassing this example, it might be better to fix it so it builds on Windows. Possibly using https://stackoverflow.com/a/26085827

danomatika:
If this has been tested and it's working on most platforms, then the last thing I would ask would be to add documentation to the readme explaining this approach and how to use it. Don't assume everyone knows what CMake is not where to get/use it, so please follow the general approach taken by the readme which attempts to explain the basics as well as the specific steps needed to get a working build environment and build.

CMakeLists.txt
eldruin: I think it would be best to refactor the code not to use `gettimeofday()`. This should be easy to do as it is only used to sleep for some time.
I opened a separate issue for this: #246

CMakeLists.txt
eldruin: Things are aligned with whitespace for readability. Actually, a few spaces are missing in this line.

eldruin:
I added the section about building with CMake to the readme now.

PR #229, opened 2018-02-10
Clean OSX target in XCode project

pierreguillot:
The commit just removes obj-c files from the OSX target in XCode project


danomatika:
Plenty of people use those targets and I plan to add a working HAL-based PdAudioUnit for desktop.

pierreguillot:
Ok sorry. As the objc/audio folder was not included I thought it was a mistake and that objective-c code was only for ios.

danomatika:
>As the objc/audio folder was not included

That's a group in Xcode (yellow folder for organization) and doesn't actually reflect the file system. A folder reference (blue folder) does.

PR #230, opened 2018-02-21
fixed the constantness of the source vector of libpd_write_array

pierreguillot:
The source vector is unchanged so it can be constant. 
I don't think it makes a big difference but it's be more explicit.

### DIFF
--- a/libpd_wrapper/z_libpd.c
+++ b/libpd_wrapper/z_libpd.c
@@ -247,7 +247,7 @@ int libpd_read_array(float *dest, const char *name, int offset, int n) {
   return 0;
 }
 
-int libpd_write_array(const char *name, int offset, float const* src, int n) {
+int libpd_write_array(const char *name, int offset, const float *src, int n) {
   sys_lock();
   MEMCPY((vec++)->w_float, *src++)
   sys_unlock();
--- a/libpd_wrapper/z_libpd.c
+++ b/libpd_wrapper/z_libpd.c
@@ -247,7 +247,7 @@ int libpd_read_array(float *dest, const char *name, int offset, int n) {
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


libpd_wrapper/z_libpd.c
```diff
@@ -247,7 +247,7 @@ int libpd_read_array(float *dest, const char *name, int offset, int n) {
   return 0;
 }
 
-int libpd_write_array(const char *name, int offset, float *src, int n) {
+int libpd_write_array(const char *name, int offset, float const* src, int n) {
```
danomatika: Can you keep the pointer formatting

    float const *src

libpd_wrapper/z_libpd.h
```diff
@@ -39,7 +39,7 @@ EXTERN int libpd_process_double(int ticks,
 EXTERN int libpd_arraysize(const char *name);
 // The parameters of the next two functions are inspired by memcpy.
 EXTERN int libpd_read_array(float *dest, const char *src, int offset, int n);
-EXTERN int libpd_write_array(const char *dest, int offset, float *src, int n);
+EXTERN int libpd_write_array(const char *dest, int offset, float const* src, int n);
```
danomatika: Also keep pointer formatting here:

    float const *src

pierreguillot:
Oops sorry I didn't noticed the formatting, I'll be more careful

danomatika:
>Oops sorry I didn't noticed the formatting, I'll be more careful

It's a small thing in this case, but you can run into subtle things by getting into the habit of always putting the pointer asterix on the type:

    int* one, two, three;
    vs
    int *one, *two, *three;

danomatika:
I'm going to wait to merge this until after the 0.11 release. It's a simple update but I don't want to worry about breaking any of the wrappers until there is time to fix them afterwards.

PR #233, opened 2018-03-20
Export symbols necessary when using libpd as shared library

eldruin:
When using libpd as a shared library on Windows there are a few symbols that are required when linking against it. These should be exported, otherwise linking is not possible.


danomatika:
Ah yes. Good catch. I've tested the build in MinGW but not usage of those particular extras.

PR #240, opened 2018-05-24
Added clang for win in makefile. Added shell/bash scripts to build  on windows

mundusnine:
Adds capability to compile with clang in mingw. On the command-line you only need to specify LLVM=True. The batch file checks if you have installed scoop(homebrew for windows), LLVM, gcc(mingw). The batch file will stop if you don't have mingw(gcc) installed. You can call the batch file with your project name and it will add the install prefix'../'+yourprojectname to the the make call with the name you passed. Msys2 needs to be installed with make.


danomatika:
A few things:

* This commit includes a bunch of libs and dlls, is this on purpose? The updated naming for the precompiled pthread on Windows is now "libwinpthread-1", which we have included already in the `libs` folders. Also, if your build system includes a package manager, can't it also install pthreads?

* As we already have a working build based on a suggested Windows setup using MinGW, I see this option as an alternative and therefore I think the script/batch file naming should reflect this. Could it be named something more like "mingw32_build_scoop.bat" and (possibly) "mingw64_build_scoop.bat" to match the existing mingw Csharp build scripts?

* I'm unfamiliar with scoop and this overall approach. I'd suggest writing up some documentation and including it under a subsection in the Windows section in the readme.

* Instead of modifying the Makefile directly, couldn't you use overrides in the batch file when calling make? We can some more options, if required. I'm hesitant to hardcode the clang version (ie. 64 bit only, etc) when it could be supplied externally. Perhaps the best solution is to simply check if CC is set, if not, then set it to "gcc" instead of the current "gcc" default we have now. This would allow detecting the system CC. **This seems to me the important point in this PR.**

Last, as far as I can tell, most of this PR is really a batch file which automates setting up a build environment for a specific build setup. Couldn't this be accomplished with a set of commandline install commands in the readme, as we have with <https://github.com/libpd/libpd#windows>? I appreciate the effort in making something easy to use, but it also becomes yet another tool to maintain if/when people start to rely on it. As such, I could see it living in a separate repository with instructions on how to use it. Our focus is really on the main Makefile, as much as possible, and it's working quite well on Windows as it is.

danomatika:
I updated the Makefile so CC is only set conditionally. This should allow using the existing env var or make command overrides. Also, you could specify extra flags outside of the makefile using the ADDITIONAL_CFLAGS var:

    make CC="clang -target x86_64-w64-mingw32" ADDITIONAL_CFLAGS="-MJ $(@).json"

windows-build.sh
```diff
@@ -0,0 +1,10 @@
+export PATH=$1:$2:$PATH
+
+cd ./
+
+find . -name '*.o' -delete
```
danomatika: Does `make clean` provide this?

danomatika:
Closing this due to lack of activity.

PR #241, opened 2018-05-25
update makefile for Windows to allow externals to load properly

Spacechild1:
right now, MinGW builds a "libpd.dll" on Windows. unfortunately, Pd externals in Windows are implicitly linked against a "pd.dll" and therefore either
A) simply won't load or 
B) crash the app if a "pd.dll" is in your system path because then Pd gets loaded twice, leading to all sorts of mayhem.

by changing the makefile to build a "pd.dll", externals will load fine. it's possible to keep the name of "libpd.lib" (and "libpd.def"), but I don't know if it's a good idea:
on the one hand it's a bit inconsistent for the import library to have a different name than the DLL, 
on the other hand you can simply do  "-lpd" with GCC on all systems.
I made two makefile variables for those files so it's easy to change them.
A third solution would be to output a pd.lib and simply make a copy named libpd.a. (this is what I actually did for the Pd build system so that we can link the extra externals with "-lpd" on all platforms).

Note: MinGW build scripts for user apps which directly link against "libpd.dll" would have to be adapted either way.

I also thought, "-static-libgcc" should go the MINGW_LDFLAGS, or is there a reason to only do it for C# builds?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ samples/csharp/**/obj/*
 samples/csharp/**/bin/*
 csharp/obj/*
 csharp/bin/*
+libs/pd.*
 libs/libpd.*
 libs/libpdnative.*
 libs/libpdcsharp.*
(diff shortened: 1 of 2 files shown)


Spacechild1:
I helped a guy from the Pd list who wasn't able to load externals with his libpd project (build with Visual Studio) and this fixed his issues. (I also tested it.)

PR #250, opened 2018-09-24
Fix CMakeLists.txt to be able to cross compile windows library using mingw on linux

diegov:
With the following change I was able to cross-compile libpd using cmake + mingw, under linux, to produce windows binaries, using a cmake toolchain file with the following:

```
SET(CMAKE_SYSTEM_NAME Windows)

SET(CMAKE_C_COMPILER x86_64-w64-mingw32-gcc)
SET(CMAKE_CXX_COMPILER x86_64-w64-mingw32-g++)
SET(CMAKE_RC_COMPILER x86_64-w64-mingw32-windres)

SET(CMAKE_FIND_ROOT_PATH  /usr/x86_64-w64-mingw32 )

set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```

I don't think it's worth including the toolchain file in the libpd sources though since paths and binary names might vary between distributions. But the cmake change is applicable to all mingw builds.

I tested the result under wine and was able to load and play a patch normally in a simple SDL application, feeding libpd's output to the SDL audio module.


danomatika:
Cool. Can you add the toolchain example to the Cmake section of the readme?

diegov:
Sure can, I'll update the PR later today with the doc changes.

diegov:
I've updated README.md, let me know if it's ok as is or if I need to add anything else.

PR #258, opened 2018-12-20
Add test patch loading test

eldruin:
This adds a basic patch loading test for two example patch files if building with a new option `-DBUILD_TESTING:BOOL=ON`


danomatika:
Couldn't we simply modify the existing C++ pdtest project to return 1on if the patch isn't loaded as use that as the test? Also, if we integrate the test project, I'd prefer if it was within the `samples/cpp` folder instead.

eldruin:
Yes, that would also be possible.

danomatika:
As of c33f999, pdtest returns -1 if `libpd_openfile()1` fails. I this should work for basic land tests.

As for integrating this into the CMake build, I think the extra C++ sample is unnecessary as:
1. pdtest can do this
2. the current PR requires a C++ compiler *only* for the C++ load test

Closing this for now. If you rework this, please create a new PR.

PR #267, opened 2019-05-26
fixed zip path traversal vulnerability

### LINKED ISSUE #266: Google warning: org.puredata.core.utils.IoUtils.extractZipResource
I received this from google:

> Zip files can contain an entry (file or directory) having path traversal characters (“../”) in its name. If developers unzip such zip file entries without validating their name, it can potentially cause a path traversal attack, leading to writes in arbitrary directories or even overwriting the files in the app's private folders.
> 
> We recommend fixing this issue in your app by checking if canonical paths to unzipped files are underneath an expected directory. Specifically, before using a File object created using the return value of ZipEntry's getName() method, always check if the return value of File.GetCanonicalPath() belongs to the intended directory path. For example:
> 
> ~~~
> InputStream is = new InputStream(untrustedFileName);
> ZipInputStream zis = new ZipInputStream(new BufferedInputStream(is));
> while((ZipEntry ze = zis.getNextEntry()) != null) {
>   File f = new File(DIR, ze.getName());
>   String canonicalPath = f.getCanonicalPath();
>   if (!canonicalPath.startsWith(DIR)) {
>     // SecurityException
>   }
>   // Finish unzipping…
> }
> ~~~

Can you solve this problem?

rpattabi:
Fixes #266

## Testing

Manually tested copy of `extractZipResource` static method in our app. Yet to verify if Google Play console still throws the security alert.

### DIFF
--- a/java/org/puredata/core/utils/IoUtils.java
+++ b/java/org/puredata/core/utils/IoUtils.java
@@ -80,14 +80,15 @@ public class IoUtils {
     List<File> files = new ArrayList<File>();
     ZipEntry entry;
     directory.mkdirs();
+    String canonicalRoot = directory.getCanonicalPath();
     while ((entry = zin.getNextEntry()) != null) {
       File file = new File(directory, entry.getName());
-      if (!file.getCanonicalPath().startsWith(directory.getCanonicalPath())) {
+      if (!file.getCanonicalPath().startsWith(canonicalRoot)) {
         // Ref: https://support.google.com/faqs/answer/9294009
         throw new SecurityException("Zip Path Traversal Vulnerability: " +
                     "Zip entry " + entry.getName() + " has " + 
                     file.getCanonicalPath() + " as its target path. " + 
-                    "But it should not be outside target dir " + directory.getCanonicalPath());
+                    "But it should not be outside target dir " + canonicalRoot);
       }
       
       files.add(file);
--- a/java/org/puredata/core/utils/IoUtils.java
+++ b/java/org/puredata/core/utils/IoUtils.java
@@ -57,6 +57,7 @@ public class IoUtils {
    * @param directory target directory
    * @return list of files that were unpacked, not including files that existed before
    * @throws IOException
+   * @throws SecurityException if zip resource has Path Traversal Vulnerability
    */
   public static List<File> extractZipResource(InputStream in, File directory) throws IOException {
     return extractZipResource(in, directory, false);
@@ -71,6 +72,7 @@ public class IoUtils {
    * @return list of files that were unpacked (if overwrite is false, this list won't include files
    *         that existed before)
    * @throws IOException
+   * @throws SecurityException if zip resource has Path Traversal Vulnerability
    */
   public static List<File> extractZipResource(InputStream in, File directory, boolean overwrite)
       throws IOException {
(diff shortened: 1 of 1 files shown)


danomatika:
I've not done any Android development personally. Does throwing an exception here crash or halt the application? If this is more of an edge case, I'd rather just swallow the error and return without doing anything or throw up a popup warning the user directly.

rpattabi:
If the app does not handle the SecurityException, indeed it will crash or halt the application. It is better than silently swallowing the error leaving the application in invalid state.

Android provides crash stacktrace in the log developers are much familiar with. As may be common with iOS, crash at the users end also provides developers a stacktrace (Google Play Console / crashlytics / etc).

Showing a popup warning or swallowing of the exception or any other handling is to be done by the callers of this static method, i.e. the app code. This decision should not be at the library code.

java/org/puredata/core/utils/IoUtils.java
```diff
@@ -82,6 +82,14 @@ public static File extractResource(InputStream in, String filename, File directo
     directory.mkdirs();
```
joebowbeer:
```suggestion
    directory.mkdirs();
    File canonicalRoot = directory.getCanonicalPath();
```

java/org/puredata/core/utils/IoUtils.java
```diff
@@ -82,6 +82,14 @@ public static File extractResource(InputStream in, String filename, File directo
...
     while ((entry = zin.getNextEntry()) != null) {
       File file = new File(directory, entry.getName());
+      if (!file.getCanonicalPath().startsWith(directory.getCanonicalPath())) {
```
joebowbeer:
```suggestion
      if (!file.getCanonicalPath().startsWith(canonicalRoot)) {
```

java/org/puredata/core/utils/IoUtils.java
```diff
@@ -82,6 +82,14 @@ public static File extractResource(InputStream in, String filename, File directo
...
+        // Ref: https://support.google.com/faqs/answer/9294009
+        throw new SecurityException("Zip Path Traversal Vulnerability: " +
+                    "Zip entry " + entry.getName() + " has " + 
+                    file.getCanonicalPath() + " as its target path. " + 
+                    "But it should not be outside target dir " + directory.getCanonicalPath());
```
joebowbeer:
```suggestion
                    "But it should not be outside target dir " + canonicalRoot);
```

java/org/puredata/core/utils/IoUtils.java
rpattabi: Good idea. One correction: `getCanonicalPath()` returns `String`.

```suggestion
    directory.mkdirs();
    String canonicalRoot = directory.getCanonicalPath();
```

danomatika:
>Android provides crash stacktrace in the log developers are much familiar with. As may be common with iOS, crash at the users end also provides developers a stacktrace (Google Play Console / crashlytics / etc).

Ok, cool. On iOS, there isn't a direct mechanism for users to see what's going on. The crashlogs are pretty deep in the settings and really only meant for developers.

>Showing a popup warning or swallowing of the exception or any other handling is to be done by the callers of this static method, i.e. the app code. 

I would then suggest making it clear, via documentation comments on the function, that this exception is a possibility and should be handled by the calling code. I'd be happy *not* to receive any future issues on this complaining about a "surprise exception." :)

rpattabi:
I agree that documenting unchecked exceptions is optional. However, I would prefer SecurityException in this case over IOException, as it conveys the problem directly.

danomatika:
Thanks!

PR #272, opened 2019-10-19
Add virtual destructors to avoid warning messages in Xcode

jim-lofelt:


danomatika:
Can you move the destructors to just after the constructors? You can also remove the comments as they are unnecessary.

danomatika:
Nevermind, I got it. Thank you.

jim-lofelt:
Thanks!

Issue #278, opened 2019-12-04
submodule from github


umlaeute:
minor nitpick, but why is the "pure-data" submodule imported from sourceforge and not from github?

i find sf to be unreliable, and it seems that the pd repo there isn't updated as regularily as the one on github...

danomatika:
It's the canonical source and we weren't sure about how the source forge <-> GitHub mirror integration was going to work with upstream the last time I updated it.

umlaeute:
yes, i thought so.

in the meantime i *think* it has become clearer, that SourceForge is no longer the "canonical source", and github is the primary public repository of Pure Data.

so i opened this ticket to trigger a re-evaluation of the canonical source :-)

PR #279, opened 2019-12-05
small build fixes

umlaeute:
apart from the `.gitmodules` switch to github (which you might want to leave out if you don't agree with my "arguing" in #278 ), this PR contains a few minor build-system fixes:

- remove bashisms from various `Makefile`: 

  the POSIX-conformant way to check for string-equality uses a *single* `=` (two equal signs as in `==` is only valid in bash).

  even better would be to just use `make` conditionals....

- slight fixup in the main `Makefile`
  that moves the `-I./pure-data/src` flag *after* the include-flags for "libpd_wrapper/"; this makes sure to prefer files from "libpd_wrapper/" over files of the same name found in "puredata/src/" (e.g. when experimenting with a pure-data branch that includes libpd...

  if there are no files of the same-name in pure-data/src/ and libpd_wrapper/ (because pure-data has not included libpd), this change won't have any effect.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+.git*      	export-ignore
+.travis.yml	export-ignore
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = git://github.com/nettoyeurny/opensl_stream.git
 [submodule "pure-data"]
 	path = pure-data
-	url = git://git.code.sf.net/p/pure-data/pure-data
+	url = git://github.com/pure-data/pure-data
--- a/Makefile
+++ b/Makefile
@@ -189,8 +189,10 @@ PDJAVA_JAR = libs/libpd.jar
 PDJAVA_SRC = libs/libpd-sources.jar
 PDJAVA_DOC = javadoc
 
-CFLAGS = -DPD -DHAVE_UNISTD_H -DUSEAPI_DUMMY -I./pure-data/src \
-         -I./libpd_wrapper -I./libpd_wrapper/util $(PLATFORM_CFLAGS) \
+CFLAGS = -DPD -DHAVE_UNISTD_H -DUSEAPI_DUMMY \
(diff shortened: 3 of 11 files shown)


danomatika:
>even better would be to just use make conditionals....

Examples? I tried to make this work but it was easier to do a shell line in the end.

umlaeute:
like that:

~~~diff
diff --git a/samples/c/pdtest/Makefile b/samples/c/pdtest/Makefile
index f15a663..5631724 100644
--- a/samples/c/pdtest/Makefile
+++ b/samples/c/pdtest/Makefile
@@ -27,7 +27,9 @@ CFLAGS = -I$(LIBPD_DIR)/pure-data/src -I$(LIBPD_DIR)/libpd_wrapper -O3
 
 $(TARGET): ${SRC_FILES:.c=.o} $(LIBPD)
        gcc -o $(TARGET) $^ $(LIBPD)
-       if [ "$(PLATFORM)" = "mac" ]; then mkdir -p ./libs && cp $(LIBPD) ./libs; fi
+ifeq ($(PLATFORM), mac)
+       mkdir -p ./libs && cp $(LIBPD) ./libs
+endif
 
 $(LIBPD):
        cd $(LIBPD_DIR) && make
@@ -37,5 +39,7 @@ clean:
 
 clobber: clean
        rm -f $(TARGET)
-       if [ "$(PLATFORM)" = "mac" ]; then rm -rf ./libs; fi
+ifeq ($(PLATFORM), mac)
+       rm -rf ./libs
+endif
        cd $(LIBPD_DIR) && make clobber
~~~

danomatika:
Ah, very nice. That hadn't occurred to me as I thought it would break the target indentation.

danomatika:
Ok. Updated as suggested: cfb98168899d65aac4beba7437533845b7c6f3e1

umlaeute:
i just looked it up in pd-lib-builder :-)

PR #280, opened 2019-12-05
C header docs update

danomatika:
This PR is essentially a port of the C API documentation into the main header files as comments. This is not meant as a replacement for the C API wiki page, but as minimal coverage for IDE usage as well as the *proposed* inclusion of the low-level C wrapper into the Pure Data repo.

I attempted to distill the important info for functions or function pointers into a minimal commenting style for compactness. I also expanded basic overview coverage for all headers, internal or not.

Header copyright has also be updated to reflect those files with major changes from their original form over the years. Let me know if this makes sense.

### DIFF
--- a/libpd_wrapper/util/z_queued.h
+++ b/libpd_wrapper/util/z_queued.h
@@ -19,55 +19,55 @@ extern "C"
 #endif
 
 /// set the queued print receiver hook, NULL by default
-/// note: avoid calling this while DSP is running
+/// note: do not call this while DSP is running
 EXTERN void libpd_set_queued_printhook(const t_libpd_printhook hook);
 
 /// set the queued bang receiver hook, NULL by default
-/// note: avoid calling this while DSP is running
+/// note: do not call this while DSP is running
 EXTERN void libpd_set_queued_banghook(const t_libpd_banghook hook);
 
 /// set the queued float receiver hook, NULL by default
-/// note: avoid calling this while DSP is running
+/// note: do not call this while DSP is running
 EXTERN void libpd_set_queued_floathook(const t_libpd_floathook hook);
 
 /// set the queued symbol receiver hook, NULL by default
-/// note: avoid calling this while DSP is running
+/// note: do not call this while DSP is running
 EXTERN void libpd_set_queued_symbolhook(const t_libpd_symbolhook hook);
 
 /// set the queued list receiver hook, NULL by default
-/// note: avoid calling this while DSP is running
+/// note: do not call this while DSP is running
 EXTERN void libpd_set_queued_listhook(const t_libpd_listhook hook);
 
 /// set the queued typed message receiver hook, NULL by default
-/// note: avoid calling this while DSP is running
+/// note: do not call this while DSP is running
 EXTERN void libpd_set_queued_messagehook(const t_libpd_messagehook hook);
 
 /// set the queued MIDI note on hook, NULL by default
-/// note: avoid calling this while DSP is running
+/// note: do not call this while DSP is running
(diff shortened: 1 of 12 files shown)


libpd_wrapper/util/z_queued.h
```diff
@@ -18,24 +18,71 @@ extern "C"
 {
 #endif
 
+/// set the queued print receiver hook, NULL by default
+/// note: avoid calling this while DSP is running
```
nettoyeurny: "Avoid calling" sounds a bit fuzzy. "Do not call" seems clearer.

libpd_wrapper/z_hooks.c
```diff
@@ -10,6 +11,7 @@
  
 #include "z_hooks.h"
 
+// libpd_printhook is sys_printhook
```
nettoyeurny: This seems misleading. I think the point is that there is no libpd_printhook because sys_printhook already serves this purpose.

libpd_wrapper/z_libpd.c
```diff
@@ -39,7 +40,7 @@ int sys_startgui(const char *libdir);
 void sys_stopgui(void);
 int sys_pollgui(void);
 
-// (optional) built in pd externals setup functions
+// (optional) built in pd core externals setup functions
```
nettoyeurny: built-in?

libpd_wrapper/z_libpd.h
```diff
@@ -18,97 +19,416 @@ extern "C"
 
 #include "m_pd.h"
 
+/* initializing pd */
+
+/// initialize libpd, it is safe to call this more than once
```
nettoyeurny: Nit: s/, it is/;/

libpd_wrapper/z_libpd.h
```diff
@@ -18,97 +19,416 @@ extern "C"
...
+/// returns 0 on success or -1 if libpd was already initialized
+/// note: sets SIGFPE handler to keep bad pd patches from crashing due to divide
+///       by 0, set any custom handling after calling this function
 EXTERN int libpd_init(void);
+
+/// clear the libpd search path for abstractions and externals
```
nettoyeurny: s/clear/clears/ for consistency

This applies here and below.

libpd_wrapper/util/z_queued.h
danomatika: Done.

libpd_wrapper/z_libpd.c
danomatika: Done.

libpd_wrapper/z_hooks.c
danomatika: Removed. This info is already in the header anyway.

libpd_wrapper/z_libpd.h
danomatika: Done.

libpd_wrapper/z_libpd.h
danomatika: I'd need to change almost every entry. So far the convention is the function will "do" something and it then "returns" some value. Maybe it looks sloppy. It could changed to "return" instead which would then make it trivial to convert the commenting into real Doxygen, if someone wants to.

danomatika: Ok, I've done a little refactoring so the main description for getters is "get ..." or "check if..." then "returns" is only used when the return range is important/non-obvious.

danomatika:
Also, thanks for the detailed review. We both know this will be "set in stone" once integrated into Pd vanilla, so it's good to get it right the first time.

PR #281, opened 2019-12-09
added array exists and resize

danomatika:
This PR adds a couple functions to the C API to check if an array exists by name and to resize a given array.

Also, as garray_resize_long() takes a long argument, this makes me think if we should consider converting sizes argument types to size_t instead of int...


danomatika:
This is also a partial response to #80.

danomatika:
Of course, perhaps `libpd_arrayexists()` is not really needed as `libpd_arraysize() != -1` suffices.

PR #282, opened 2019-12-12
[WIP] Feature/multi hooks

danomatika:
This PR helps rectify one of the issues with libpd's current implementation: you can send to specific instances but multi-instance receiving is to the same, shared hooks.

* [x] per-instance print hooks (involves changes to pd core)
* [x] per-instance float, symbol, & message hooks; working via libpdreceiver object
* [x] per-instance MIDI out hooks (~~involves changes pd to core~~ added to libpd itself)

There are also the layers and various wrappers which may also need to be updated:
* [x] print concatenator
* [x] queued layer (required API change to where `libpd_queued_init()` must be called before setting any queued hooks)
* [x] C++
* [ ] ~Obj-C~ later
* [ ] ~Java?~ later
* [x] C#? see #222
* [ ] ~Python?~ later

See [companion pure-data PR](https://github.com/pure-data/pure-data/pull/836). (merged)


danomatika:
Ok, I think I have all the hooks working per-instance.

@pierreguillot Can you check the `pdtest_threads` C example? I'm seeing a runtime error so far:

~~~
* thread #6, stop reason = EXC_BAD_ACCESS (code=1, address=0x38)
  * frame #0: 0x00000001001b60c0 libpd.dylib`sys_lock at s_inter.c:1679:25
    frame #1: 0x0000000100207384 libpd.dylib`libpdinstance_new at z_hooks.c:45:3
    frame #2: 0x00000001002095c4 libpd.dylib`libpd_new_instance at z_libpd.c:614:3
    frame #3: 0x000000010000312c pdtest_thread`libpd_instance_doinit(inst=0x000000016fdfd608) at pdtest_thread.c:35:18
    frame #4: 0x000000018b577878 libsystem_pthread.dylib`_pthread_start + 320
~~~

danomatika:
Has anyone had a chance to test this further?

Ant1r:
I'll try to find some time today for testing this.

pierreguillot:
Sorry, I'll test it this week !

Ant1r:
At least `pdtest_multi` works as expected here on Linux.
Though I don't understand why you changed this in pdtest_multi.c:
```diff
 void pdprint1(const char *s) {
-  printf("pd1: %s", s);
+  printf("%s", s);
 }
 
 void pdprint2(const char *s) {
-  printf("pd2: %s", s);
+  printf("%s", s);
 }
```
IMO the prefix was making it more obvious that per-instance-printhook works properly.

I'll to test this branch in real life situation (e.g ofxPd on Android) soon.

Ant1r:
OTOH, `pdtest_thread` sample doesn't work anymore in this branch here.
Here's the backtrace, I didn't dig further (I compiled `libpd` with `DEBUG=true  MULTI=true THREAD=true` flags):

```gdb
Starting program: /home/arsene/Pd/libpd/samples/c/pdtest_thread/pdtest_thread test.pd .
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
pd 0.52.0test3
bonk version 1.5
fiddle version 1.1 TEST4
pd~ version 0.54
pique 0.1 for PD version 23
sigmund~ version 0.07
[New Thread 0x7ffff6f78700 (LWP 4603)]
[New Thread 0x7ffff6777700 (LWP 4604)]
[New Thread 0x7ffff5f76700 (LWP 4605)]
[New Thread 0x7ffff5775700 (LWP 4606)]
[New Thread 0x7ffff4f74700 (LWP 4607)]
[New Thread 0x7fffdffff700 (LWP 4609)]
[New Thread 0x7fffdf7fe700 (LWP 4608)]
[New Thread 0x7fffdeffd700 (LWP 4610)]

Thread 4 "pdtest_thread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff5f76700 (LWP 4605)]
0x00007ffff7b2f787 in sys_lock () at pure-data/src/s_inter.c:1700
1700	    pthread_mutex_lock(&INTER->i_mutex);
(gdb) bt
#0  0x00007ffff7b2f787 in sys_lock () at pure-data/src/s_inter.c:1700
#1  0x00007ffff7b7d851 in libpdinstance_new () at libpd_wrapper/z_hooks.c:45
#2  0x00007ffff7b7f6c9 in libpd_new_instance () at libpd_wrapper/z_libpd.c:614
#3  0x0000000000401029 in libpd_instance_doinit (inst=0x7fffffffbc00) at pdtest_thread.c:35
#4  0x00007ffff78576ba in start_thread (arg=0x7ffff5f76700) at pthread_create.c:333
#5  0x00007ffff758d51d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
(gdb) 

```

danomatika:
Yeah, I wasn’t able to debug the thread test crash, sonI was hoping someone else might have some insight.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

pierreguillot:
I've been debugging the pdmulti_test on macOS using Xcode and it seems that when calling `libpdinstance_new()`, `libpd_this = NULL` and `pd_this = NULL` so  [sys_lock()](https://github.com/libpd/libpd/pull/282/files#diff-f724f0817e35f4a9d15875090bab2b1ff311f38f5feef52a82e2a5700a221342R45) crashes. I don't understand yet why `libpd_this = NULL` and `pd_this = NULL` but I'll continue my investigation.

pierreguillot:
But the problem seems to be the same with pdthread_test.

pierreguillot:
I might be wrong but it seems that on this branch, the libpd files are duplicated between this repository and the pure-data repository that embeds its own version of libpd. This could introduce issues and perhaps that's why `pd_this` is NULL. In my opinion, it would be good to fix this.

Ant1r:
> the libpd files are duplicated between this repository and the pure-data repository

I don't think this is a problem: while compiling libpd from this repo, you can notice that pure-data/z_* files are NOT included:
```
cc -o libs/libpd.so pure-data/src/d_arithmetic.o pure-data/src/d_array.o pure-data/src/d_ctl.o pure-data/src/d_dac.o pure-data/src/d_delay.o pure-data/src/d_fft.o pure-data/src/d_fft_fftsg.o pure-data/src/d_filter.o pure-data/src/d_global.o pure-data/src/d_math.o pure-data/src/d_misc.o pure-data/src/d_osc.o pure-data/src/d_resample.o pure-data/src/d_soundfile.o pure-data/src/d_soundfile_aiff.o pure-data/src/d_soundfile_caf.o pure-data/src/d_soundfile_next.o pure-data/src/d_soundfile_wave.o pure-data/src/d_ugen.o pure-data/src/g_all_guis.o pure-data/src/g_array.o pure-data/src/g_bang.o pure-data/src/g_canvas.o pure-data/src/g_clone.o pure-data/src/g_editor.o pure-data/src/g_editor_extras.o pure-data/src/g_graph.o pure-data/src/g_guiconnect.o pure-data/src/g_hdial.o pure-data/src/g_hslider.o pure-data/src/g_io.o pure-data/src/g_mycanvas.o pure-data/src/g_numbox.o pure-data/src/g_readwrite.o pure-data/src/g_rtext.o pure-data/src/g_scalar.o pure-data/src/g_template.o pure-data/src/g_text.o pure-data/src/g_toggle.o pure-data/src/g_traversal.o pure-data/src/g_undo.o pure-data/src/g_vdial.o pure-data/src/g_vslider.o pure-data/src/g_vumeter.o pure-data/src/m_atom.o pure-data/src/m_binbuf.o pure-data/src/m_class.o pure-data/src/m_conf.o pure-data/src/m_glob.o pure-data/src/m_memory.o pure-data/src/m_obj.o pure-data/src/m_pd.o pure-data/src/m_sched.o pure-data/src/s_audio.o pure-data/src/s_audio_dummy.o pure-data/src/s_inter.o pure-data/src/s_loader.o pure-data/src/s_main.o pure-data/src/s_net.o pure-data/src/s_path.o pure-data/src/s_print.o pure-data/src/s_utf8.o pure-data/src/x_acoustics.o pure-data/src/x_arithmetic.o pure-data/src/x_array.o pure-data/src/x_connective.o pure-data/src/x_file.o pure-data/src/x_gui.o pure-data/src/x_interface.o pure-data/src/x_list.o pure-data/src/x_midi.o pure-data/src/x_misc.o pure-data/src/x_net.o pure-data/src/x_scalar.o pure-data/src/x_text.o pure-data/src/x_time.o pure-data/src/x_vexp.o pure-data/src/x_vexp_if.o pure-data/src/x_vexp_fun.o libpd_wrapper/s_libpdmidi.o libpd_wrapper/x_libpdreceive.o libpd_wrapper/z_hooks.o libpd_wrapper/z_libpd.o libpd_wrapper/util/z_print_util.o libpd_wrapper/util/z_queued.o libpd_wrapper/util/ringbuffer.o -shared -Wl,-Bsymbolic -ldl  -lm -lpthread
```
Only libpd_wrapper/ ones are included.

----------

> pd_this = NULL

yeah, I followed the same path.
it seems that in `z_hooks.c/libpdinstance_new()`, `sys_lock()` is called before `m_class.c/pdinstance_new()`, which is the one that initializes `pd_this` then calls `s_inter.c/s_inter_newpdinstance()` which in turn initializes the mutex. 

Hence the mutex isn't initialized yet when `sys_lock()` is called.

I tried to change it like this:
```C
t_libpdinstance *libpdinstance_new(void) {
  t_libpdinstance *x = (t_libpdinstance *)calloc(1, sizeof(t_libpdinstance));
  x->hooks = libpdhooks_new();
  x->pd = pdinstance_new();// increments pd_ninstances
  libpd_this = x;
  /* now the instance is initialized, we can call sys_lock(): */
  sys_lock();
  libpd_instances = (t_libpdinstance **)resizebytes(libpd_instances,
        pd_ninstances * sizeof(*libpd_instances),
        (pd_ninstances+1) * sizeof(*libpd_instances));
  libpd_instances[pd_ninstances] = x;
  sys_unlock();
  return x;
}
```

It seems I can go one step further thanks to this modification, but then it crashes for other reasons:
- first it crashed because of a call to sys_vgui:
```
 *** Error in `./pdtest_thread': realloc(): invalid pointer: 0x00007fad240000b8 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777f5)[0x7fad45ffa7f5]
/lib/x86_64-linux-gnu/libc.so.6(realloc+0x358)[0x7fad46007a68]
../../../libs/libpd.so(+0xbfbaa)[0x7fad46629baa]
../../../libs/libpd.so(sys_vgui+0x1b9)[0x7fad46629f0a]
../../../libs/libpd.so(sched_set_using_audio+0xa3)[0x7fad4662601f]
../../../libs/libpd.so(libpd_init_audio+0x8f)[0x7fad4667a4f0]
```
- then I commented the last line of sys_set_audio_settings() which calls sys_vgui(), and I got another crash related to a bad address while calling free():
```
*** Error in `./pdtest_thread': free(): invalid pointer: 0x00007f49500000e8 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777f5)[0x7f49698ac7f5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8038a)[0x7f49698b538a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7f49698b958c]
../../../libs/libpd.so(freebytes+0x2b)[0x7f4969ed3f59]
../../../libs/libpd.so(sys_setchsr+0xcc)[0x7f4969ed886d]
../../../libs/libpd.so(sys_reopen_audio+0x90)[0x7f4969ed938c]
../../../libs/libpd.so(libpd_init_audio+0x94)[0x7f4969f2c4cb]
```

Generally, I wonder if libpd should have its own mutex, to protect the creation and destruction of instances, since `sys_lock()` is actually using a per-instance mutex.

danomatika:
>Generally, I wonder if libpd should have its own mutex, to protect the creation and destruction of instances, since sys_lock() is actually using a per-instance mutex.

We didn't need one before, so I was hoping to avoid adding one. Perhaps we could use one just for accessing the libpd-specific instance stuff I added?

pierreguillot:
The problem is not necessarily with the source files but with the header files. It’s not clear which one is used now.

danomatika:
libpd sources are still currently those found in this repository. The makefile doe not use those now merged with the pd core sources.

My plan is to first get this PR working, then remove the libpd sources for this repo as it easier to modify the libpd sources here. Once everything is working I can make a PR to pure-data.

danomatika:
As a test, delete/rename those in pure-data/src.

Ant1r:
> I wonder if libpd should have its own mutex

anyway, I tried to protect `libpdinstance_new()` with a new `libpd_[un]lock()` using a global `libpd_mutex`, and it still crashes at the same stage (currently in `sys_setchsr()`).

danomatika:
@Spacechild1 Maybe you have some insight?

pierreguillot:
I'll do further tests tomorrow. I suspect that libpd files and/or pd files have been modified when copying libpd files to the  Pd repository and now this repository doesn't support the changes.

Ant1r:
One step beyond:
I discovered that `libpd_set_instance()` was actually wrongly setting `pd_this`!
It can be seen with tracing `pdtest_thread.c::libpd_instance_doinit()`:
```C
static void* libpd_instance_doinit(t_libpd_instance* inst)
{
    printf("libpd_instance_doinit() start\n");
    inst->l_pd = libpd_new_instance();
    printf("libpd_instance_doinit() libpd_new_instance pd_instance= %p\n", inst->l_pd);
    libpd_set_instance(inst->l_pd);
    assert(inst->l_pd && "pd instance can't be allocated.");
    printf("libpd_instance_doinit() after libpd_set_instance() pd_this= %p\n", pd_this);
    libpd_init_audio((int)inst->l_ninputs, (int)inst->l_noutputs, (int)inst->l_samplerate);
    return NULL;
}
```

which returns for example:
```
libpd_instance_doinit() libpd_new_instance pd_instance= 0x7f1d54000950
libpd_instance_doinit() after libpd_set_instance() pd_this= 0x7f1d54000078

```

If I comment the `libpd_set_instance()` line (`pd_this` and `libpd_this` are already initialized by `libpd_new_instance()` anyway), the 2 prints return the same address, and it doesn't crash anymore when calling `libpd_init_audio()`.

But then it crashes later, since `libpd_set_instance()` is called at other places (now it crashes in `pdtest_thread.c::libpd_instance_doopen()`).

So I think there's a bug in `libpd_set_instance()`.

Ant1r:
Maybe the bug comes from:
```C
void libpd_set_instance(t_pdinstance *p) {
#ifdef PDINSTANCE
  libpdinstance_set(libpd_instances[p->pd_instanceno]);
#endif
}
```
`p->pd_instanceno` is the index of the instance in `pd_instances[]`, not in `libpd_instances[]`.

danomatika:
My idea was that there should be the same number of libpd instances as of instances, although of course this breaks if people don’t use the libpd instance function wrappers…

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Ant1r:
that's it!
there's an offset of 1 between pd_instances[] and libpd_instances[].
It might come from the fact `libpd_init()` already creates a "dummy" pd_instance.

If I change 
```C 
libpdinstance_set(libpd_instances[p->pd_instanceno]);
``` 
to 
```C
libpdinstance_set(libpd_instances[p->pd_instanceno + 1]);
```
and the same for `libpdinstance_free()` I got the right address and most crashes disappear!

It still crashes when instances are deleted (I think).
Still have to investigate, but progress is being made ;-)

danomatika:
Ah whoops. I kind of assume pd created a default instance, then additional instances afterwards.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
> Though I don't understand why you changed this in pdtest_multi.c:
> 
> ```diff
>  void pdprint1(const char *s) {
> -  printf("pd1: %s", s);
> +  printf("%s", s);
>  }
>  
>  void pdprint2(const char *s) {
> -  printf("pd2: %s", s);
> +  printf("%s", s);
>  }
> ```
> 
> IMO the prefix was making it more obvious that per-instance-printhook works properly.

The prefix was breaking the line breaks when both instanced were printing. The real fix is to use the print concatenation layer, but I was too lazy then to enabled it.

Ant1r:
> The prefix was breaking the line breaks

ah, OK. I'll see if I can help there. What do you call "the print concatenation layer"?

> there's an offset of 1 between pd_instances[] and libpd_instances[].

Wouldn't it possible in first place to avoid the use of `t_libpdinstance` and `libpd_instances[]`?

Afterall the only reason of `t_libpdinstance` is to add the `t_libpdhooks *hooks` member; but how about adding the hooks (or the `t_libpdhooks *hooks`) to the `t_instancestuff` structure, which  already has the t_printhook member?

It seems to me that it would avoid an unnecessary complication.

danomatika:
At the time, libpd was not integrated into pure-data and I preferred to keep the libpd-specific stuff separated, at least logically.

danomatika:
>which already has the t_printhook member?

Actual this isn't true... I'm adding via https://github.com/pure-data/pure-data/pull/836/files since this is the only change, so far, needed in the pure-data repo.

Ant1r:
> Actual this isn't true... I'm adding via [pure-data/pure-data#836](https://github.com/pure-data/pure-data/pull/836)

Ah yes! 
So don't you think `t_libpdhooks *hooks` could be added to STUFF, since STUFF already has to be hacked?

danomatika:
Again, it’s only used by libpd and I’d prefer to keep things like this modular. We could of course add everything to STUFF but then where is the limit? The message hooks don’t need to be accessible to anything outside of libpd, only the libpd receiver. The MIDI hooks are called inside libpds outmidi interface which just implements those function stubs from s_midi.c which isn’t compiled into libpd. I initially looked into adding the MIDI hooks to pure-data but there isn’t really a need to do that as, again, we implement the MIDI dummy layer separately (modular).

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
If I think about it, it makes more sense to add an implementation-specific void pointer to STUFF, then we can stick what we need to the underlying instance struct without codifying libpd-specific strictest outside of libpd headers. This would be a compromise between adding to STUFF and keeping specifics separate. I can try this later.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Spacechild1:
> it makes more sense to add an implementation-specific void pointer to STUFF

I think this is a great idea! Library APIs often allow clients to pass some kind of opaque user data per instance/context.

This way you don't have to duplicate the whole per-instance layer in libpd - which indeed feels a bit redundant.

Ant1r:
> We could of course add everything to STUFF but then where is the limit?

Well, the whole `t_instancestuff` purpose is to allow multiple instancing. If we only add one `t_libpdhooks *hooks` member, it could be NULL in normal Pd case, and initialized only in libpd context.
This would avoid the additional nesting of `t_libpdinstance` and `libpd_instances[]` and the associated bugs.

Spacechild1:
@Ant1r I think Dan's last suggestion is much better: just add an opaque pointer to `STUFF`, e.g. `void *s_context` that libpd can set to whatever it wants. This way libpd can freely add more per-instance things and Pd wouldn't need to be aware of it.

Ant1r:
Yes of course (sorry I had not read your answers yet when I wrote my last message)!

If libpd has to introduce a new struc e.g `t_libpdcontext`, I think it would be good to add an additional opaque pointer, for possible needs of the application developer.
Maybe something like:
```C
typedef struct _libpdcontext {
  t_libpdhooks *hooks;
  void *usercontext; /* can be used freely by the application developer to store per-instandce data */
} t_libpdcontext;
```

As to whether the `s_context` should be added to `t_pdinstance` or `t_instancestuff`, I admit I have no clear idea...

Ant1r:
@danomatika , or @Spacechild1, would you prefer (and do you have the time) to take care of this by yourself, or would you like me to try to do it?

danomatika:
I am making a test implementation now.

danomatika:
Ok, I refactored to use pd_instance directly which simplifies everything. No more trying to keep extra pointers around or guess lock timing. I also added per-instance user data.

The C tests all work. protest_thread doesn't crash but only outputs empty lines after the usually load prints. AT least it doesn't crash anymore.

danomatika:
Also, I went with "implementation" as the conceptual name for the per-instance pointer, ie. `st_impdata`, `t_libpdimp`, etc. I know "context" was suggested but it seemed a little too generic and "instance" seems overblown for what is essentially "_libpdinstancestuff". Any further thoughts on this?

libpd_wrapper/z_libpd.h
```diff
@@ -473,31 +473,39 @@ EXTERN int libpd_poll_gui(void);
 
 /* multiple instances */
 
-/// create a new pd instance
+/// create a new pd instance and set as current
+/// note: use this in place of pdinstance_new()
 /// returns new instance or NULL when libpd is not compiled with PDINSTANCE
-EXTERN t_pdinstance *libpd_new_instance(void);
+EXTERN t_pdinstance *libpd_new_instance();
```
Spacechild1, line 479: missing `void` :-)

Spacechild1:
Yay!

`"z_hooks.h"` is a public header, right? I have been wondering about following function declarations:
```
t_libpdhooks *libpdhooks_new(void);
void libpdhooks_free(t_libpdhooks *hooks);
t_libpdimp* libpdimp_new(void);
void libpdimp_free(t_libpdimp *imp);
```
AFAICT, these are only used internally. Or is there a situation where a libpd user might need them? The only function we really need is `libpdimp_this`, I guess.

danomatika:
> `z_hooks.h` is a public header, right?

[Nope](https://github.com/libpd/libpd/blob/feature/multi-hooks/libpd_wrapper/z_hooks.h#L18)

I have been wondering about following function declarations:
> 
> ```
> t_libpdhooks *libpdhooks_new(void);
> void libpdhooks_free(t_libpdhooks *hooks);
> t_libpdimp* libpdimp_new(void);
> void libpdimp_free(t_libpdimp *imp);
> ```
> 
> AFAICT, these are only used internally.

Yep, and they should stay that way. User can use `libpd_set_instancedata()` and `libpd_get_instancedata()`.

Spacechild1:
Ah, now I see!

> Users can use `libpd_set_instancedata()` and `libpd_get_instancedata()`

Nice!

Ant1r:
Congrats @danomatika it's all OK here (once the `libpd_new_instance()` definition is fixed) !

pdtest_thread prints normally if you add a `pdprint(const char *s)` and you set the printhook. Apparently the default printhook doesn't work anymore.

> I went with "implementation" as the conceptual name for the per-instance pointer `st_impdata`

Actually is this pointer likely to be used by anything else than libpd? If the answer is no, then maybe it could be simply named like `st_libpdstuff`, to make it easier to understand when reading the source.
OTOH "implementation" is fine enough to me ;-)

Ant1r:
BTW, the `concatenated_print` and `queued` parts also have to be rewritten.
I have a working version of  `concatenated_print` here, which requires to add some per-instance stuff:
```C
typedef struct _libpdhooks {

  char *h_concatenated_printbuffer;
  int h_concatenated_printbuflen;
  t_libpd_printhook h_concatenated_printhook;
  (...)
```
Would you agree with that?

As for `queued` messaging, it would require to add a full list of `h_queued_[*]hook` members, or create a new struct to pack them in.
But that seems a bit overkill to me: why not just adding a `h_isqueued` then adapting the logic? In this case, all the `libpd_set_queued_[*]hook()` functions could be removed, in favor of a new `libpd_set_queued(int queued)` for example.

Ant1r:
I sent a proposition for the concatenated print, as #343.
Finally the added private stuff is kept opaque in `z_print_util.c`.
I'm planning to do the same for the "queued" part.

Ant1r:
the "queued" part is now multi-instance compatible in #343 .

danomatika:
Sorry, I have been busy with work this week and probably won't be able to look at this again until next week. Thanks in advance.

Without looking at the proposed changes, I would just ask to *not* break the current API if at all possible as it's used by the wrappers and user projects.

Ant1r:
OK Dan, I understand that.
The change to the API is minimal, and maybe I should bring dummy versions back of the only 2 functions I finally removed (only libpd_queued_init() and libpd_queued_release()), and mark them as deprecated.

danomatika:
Can they be wrapper functions in some way?

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Ant1r:
Well, yeah I guess `libpd_queued_init()` could simply return `libpd_init()`.

As for `libpd_queued_release()`, I think it should do nothing, since every allocated memory chunk is freed by `libpd_free_instance()`.

Ant1r:
OK, `libpd_queued_init()` and `libpd_queued_release()` are back.
I've also removed the first `t_pdinstance *pd` parameter of the `libpd_[g/s]et_instancedata()`, since it's useless.

Ant1r:
Sidenote: having duplicated files in `pure-data/src` and `libpd_wrapper` can indeed cause some problems.
For example, if `z_libpd.h` is first found in `pure-data/src`, `libpd_[g/s]et_instancedata()` functions are not declared, hence the compilation fails.

I've edited the `sample/c/*/Makefile`s to make sure `libpd_wrapper/` is searched before.

danomatika:
This branch should be working now with the new Pd 0.52-2 sources. Please update and test:

    git checkout feature/multi-hooks
    git submodule update

So far all the of C tests are working fine. I added some printfs for basic output confirmation to pdtest_thread as well. If we are good to go, then I would merge this to master and release as part of libpd 0.13. I was too late this time around, but the next version of pd can include these changes and we could remove the sources in `libpd_wrapper` in this repo.

danomatika:
I used #343 as a guide and implemented instance support for the print util and queued layers. Also, the user data set function now accepts an optional freehook which is used when clearing the instance data. The C samples are all updated and I added Antoine's `pdtest_queued`, so please test.

Ant1r:
Thanks, it works great here (appart from the `c/pdtest_queued/Makefile` to be fixed adding -lpthread for Linux)!
I like the optional data freehook!

danomatika:
> Thanks, it works great here (appart from the `c/pdtest_queued/Makefile` to be fixed adding -lpthread for Linux)!

Try changing LDFLAGS `+=` to `=` [in the Makefile](https://github.com/libpd/libpd/blob/adc73b0c4e152c8d3138c09e54416b08bfb13daa/samples/c/pdtest_thread/Makefile#L15) and let me know if it then links correctly...

danomatika:
> I like the optional data freehook!

I needed something like it internally in order to keep the queued implementation separate and it made sense to also expose the same mechanism for user data. With multiple instances, it should be easier to make sure you free all your memory.

Ant1r:
It's for pdtest_queued sample. pdtest_thread  compiles normally.

danomatika:
> It's for pdtest_queued sample. pdtest_thread compiles normally.

Ok. Should be fixed now.

danomatika:
I am more or less happy with the C & C++ implementations of this now. The other language wrappers can be updated later as it's too much work to try to do this all now.

@nettoyeurny If you have time, can you do a quick review of the changes to the C API? Once cleared, I would merge for a new release.

libpd_wrapper/x_libpdreceive.c
```diff
@@ -19,33 +18,39 @@ static t_class *libpdrec_class;
...
+  if (x->x_hooks->h_banghook)
+    (*x->x_hooks->h_banghook)(x->x_sym->s_name);
 }
 
 static void libpdrecfloat(t_libpdrec *x, t_float f) {
-  if (libpd_floathook) (*libpd_floathook)(x->x_sym->s_name, f);
-  if (libpd_doublehook) (*libpd_doublehook)(x->x_sym->s_name, f);
+  if (x->x_hooks->h_floathook)
+    (*x->x_hooks->h_floathook)(x->x_sym->s_name, f);
+  if (x->x_hooks->h_doublehook)
```
nettoyeurny, line 32: My understanding is that at most one of h_{float,double}hook will be non-null, so "else if" seems closer to the intention?

nettoyeurny (approved):
Looks great, just one tiny nit about readability :) This is a major milestone --- thank you so much for all your work!

libpd_wrapper/x_libpdreceive.c
danomatika: Probably right. If one is set the other will be NULL. I'll update this in `z_queued.c` as well.

danomatika:
@nettoyeurny Thanks for the feedback. I somehow lost track of this but am getting the release together now that pd 0.53-0 is out.

danomatika:
Thanks for all the feedback & testing. Here we go...

PR #288, opened 2020-03-01
This fixes #287

### LINKED ISSUE #287: CMake: Could not find libm
CMake could not find math library libm. Error during linking library to target libpd in [CMakeLists.txt](https://github.com/libpd/libpd/blob/master/CMakeLists.txt), line 244. Wrong call of library variable.

etk70182:
Fixing linkage of math library to target. (Issue occured on Ubuntu 18.04 with CMake 3.10.2)


danomatika:
Thanks

PR #291, opened 2020-04-13
Update podspec: requires_arc = true

amethystdeceiver:
Seems like the transition to ARC happened years ago (329c5a28af72e71d98b1a8cff923f715b8341bec), but the podspec hasn't been updated


danomatika:
Yes and thanks. None of the libpd devs use CocoaPods. I update the file list if I need to, but that's about it.

PR #306, opened 2020-07-06
Static build target

clwe:
This branch adds a static build target option to the makefile for unix and windows(mingw64) libraries.
To build the static library ```libpd.a``` use the command ```make STATIC=true```


danomatika:
Can you make the build var simply "STATIC"?

clwe:
:) ok!

danomatika:
Thanks. I tweaked this a bit to build the `libpd.a` in addition to the regular dynamic lib and it is also installed if built.

PR #307, opened 2020-09-15
Poll gui return value

millerpuckette:
have libpd_poll_gui() report what sys_poll() returned so the caller can decide whether to poll again


danomatika:
Hmm. Weird that it's 16 commits for 10 line changes.

millerpuckette:
Hmm - I assumed git would see that the last 15 are just me trying to get
may master in sync with libpd's master.  The last commit is the only one
that matters.

I'd have been happy just to delete the fork and start over from scratch but
I don't know how to delete a github project :)

M

millerpuckette:
Here's a better way, just AM this...

From f01c2ca70275cb5f70ca839677a88b996f217516 Mon Sep 17 00:00:00 2001
From: Miller Puckette <msp@ucsd.edu>
Date: Tue, 15 Sep 2020 11:01:17 -0700
Subject: [PATCH] pass sys_poll() return value on to libpd_poll_gui() caller

---
 libpd_wrapper/z_libpd.c | 6 ++++--
 libpd_wrapper/z_libpd.h | 4 +++-
 2 files changed, 7 insertions(+), 3 deletions(-)

diff --git a/libpd_wrapper/z_libpd.c b/libpd_wrapper/z_libpd.c
index 5c21b81..3eab73f 100644
--- a/libpd_wrapper/z_libpd.c
+++ b/libpd_wrapper/z_libpd.c
@@ -593,10 +593,12 @@ void libpd_stop_gui(void) {
   sys_unlock();
 }
 
-void libpd_poll_gui(void) {
+int libpd_poll_gui(void) {
+  int retval;
   sys_lock();
-  sys_pollgui();
+  retval = sys_pollgui();
   sys_unlock();
+  return (retval);
 }
 
 t_pdinstance *libpd_new_instance(void) {
diff --git a/libpd_wrapper/z_libpd.h b/libpd_wrapper/z_libpd.h
index 2dbe5f4..c36f30a 100644
--- a/libpd_wrapper/z_libpd.h
+++ b/libpd_wrapper/z_libpd.h
@@ -467,7 +467,9 @@ EXTERN void libpd_stop_gui(void);
 /// this is called automatically when using a libpd_process function,
 /// note: this also facilitates network message processing, etc so it can be
 ///       useful to call repeatedly when idle for more throughput
-EXTERN void libpd_poll_gui(void);
+///       Returns 1 if the poll found something, in which case it might be
+///       desirable to poll again, up to some reasonable limit.
+EXTERN int libpd_poll_gui(void);
 
 /* multiple instances */
 
-- 
2.26.2

danomatika:
I usually just copy the affected files out of the dir and/or hard reset or clone a fresh copy, then replace the files and make a new, clean PR.

millerpuckette:
Hmm... will give it a try - perhaps I can then force-push to github.

danomatika:
If you do the work in feature branches, there is no need to force anything. Just delete branch and make a new one.

millerpuckette:
Ha...

No, I had polluted the master branch with no way to un-pollute it that I could
see.  Anyhow I think I've now wiped the whole repo and made a new PR with only
the goods :)

M

PR #308, opened 2020-09-15
pass sys_poll() return value on to libpd_poll_gui() caller

millerpuckette:


danomatika:
Does this require a new version of Pd?

millerpuckette:
It goes back at least to 0.48, probably further.

cheers
M

millerpuckette:
P.S. I'm working on 0.51-3, which will probably be worth pulling into libpd
when it's solid.

danomatika:
Ok, cool.

PR #309, opened 2020-09-25
CMake: Keep output files under build directory

johnkeeping:
Writing libraries back into the source tree breaks any workflow that
uses separate build directories, especially using a single source tree
to compile multiple configurations or for multiple architectures.

Instead, write into the build directory, keeping the "libs" subdirectory
so that in-source builds maintain the current behaviour.


danomatika:
Is this something that should be noted in the CMake section in the README.txt?

johnkeeping:
There's nothing in the README about the location of build products at the moment, and this change makes the behaviour less surprising so I'm not sure that it's necessary to add something now.

danomatika:
I think you raise a good point that it would be a good to:

1. standardize the resulting built lib location
2. note this location

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Thanks

PR #277, opened 2019-12-04
expose double-precision access

umlaeute:
adds `double` variants for number-handling

for discussion about #276

Issue #276, opened 2019-12-04
double-precision support


umlaeute:
as raised by @danomatika in https://github.com/pure-data/pure-data/pull/807, it might be a good idea to have libpd support double-precision builds of Pd properly.

umlaeute:
afaict, libpd (the "low-level C API") will not have any problems with double-precision as is.
the `libpd_process_double` will already allow you to have full benefits of a double-precision build.
so no changes required for sample-processing!

however, the message-subsystem as exposed via libpd only offers a single-precision interface.
so you might want to add double-precision variants of the following functions, so that libpd can make *full* use of a double-precision Pd:

- `libpd_read_array`, `libpd_write_array`
- `libpd_float`
- `libpd_set_float`, `libpd_get_float`

you also might want to implement a `doublehook` (that probably should precedence over the `floathook`).

umlaeute:
the alternative is to use `t_float` instead of explicit `float` and `double` variants.

i'm leaning towards explicit variants of the various float-functions, as i guess this would allow an application to (dynamically) link against a `libpd.so` and not having to worry whether this library was compiled with double or single precision internally.

this only works as long as the application doesn't know about `t_float`.
Currently the *libpd* API doesn't mention `t_float` at all, so (it seems that) an application can chose to be oblivious of the internal sampletype of Pd.

danomatika:
Using t_float would probably also introduce all sorts of headaches in the language wrappers.

danomatika:
So I've merged #310 but I'm not able to get it working so far...

A modification of the pdtest c sample which uses `libpd_process_double()` and a test patch with `[osc~ 0.1]` and prints 1/3 is not giving me output with more decimal places. Both libpd and the program are compiled with `-DPD_FLOATSIZE=64`.

umlaeute:
i cannot reproduce.
i also modified samples/c/pdtest/pdtest.c, like this:

```C
#include <stdio.h>
#include "z_libpd.h"

void pdprint(const char *s) { ; }
void pdnoteon(int ch, int pitch, int vel) { ; }

#ifdef USE_DOUBLE
# define TYPE double
# define PROCESS libpd_process_double
#else
# define TYPE float
# define PROCESS libpd_process_float
#endif

#define print_samples(num, samples) \
        for(unsigned int i=0; i<num; i++) printf("%+f%c", samples[i], (!((i+1)%8))?'\n':'\t'); printf("\n");

int main(int argc, char **argv) {
[...]

PR #310, opened 2020-11-01
expose double-precision access

umlaeute:
i somehow lost my handle to PR #277 (so i think it's easiest to just reboot this PR from a new clone).

here's the initial description:
> this PR adds `double` variants for number-handling.
> 
> for discussion about #276

and some more:
> reviewer wishlist: @danomatika & @nettoyeurny
> the code has not been tested in real-world.
> it looks trivial enough, but the devil never sleeps...

PR updated to 541ad5ec6d74779172dbeffadfb05041827e7ecd (aka today's `master`)


danomatika:
Could reading/writing doubles to a float array have potential overruns? Maybe not if it's all through a t_float?

This also makes me wonder if there should also be a define or function to check if libpd is using floats or double.

danomatika:
I'm planning on releasing libpd 0.12 once Pd 0.51-3 is finalized as a stable release. I think it makes sense to merge this after so people can try it out for a while.

umlaeute:
> Could reading/writing doubles to a float array have potential overruns? Maybe not if it's all through a `t_float` internally?

right. we don't do any `memcpy`-like copying, but instead copy the arrays element-by-element, so there shouldn't be any issues.

there's room for optimization here (if `sizeof(t_word)==sizeof(targetfloat)` we could use a proper `memcpy`), but that would need to be evaluated against the code generated by a modern C-compiler first.


> I'm planning on releasing libpd 0.12 once Pd 0.51-3 is finalized as a stable release

cool. i'd like to update https://github.com/pure-data/pure-data/pull/519 ;-)

>  I think it makes sense to merge this after so people can try it out for a while.

but then: maybe only after that....

danomatika:
libpd 0.12 is almost done. There are some issues with the Python wrapper (2 vs 3) then it’s ready.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
> There are some issues with the Python wrapper (2 vs 3) then it’s ready.

sidenote: i think the default should by Python3 (Python2 has seen EOL almost a year ago, and it's getting increasingly harder to get it - and rightly so)

https://github.com/libpd/libpd/blob/0880cbf9a575c926894d48957e9a170f9dc2e4d0/python/Makefile#L1-L2

danomatika:
Ok, great. I only use Python now and then. It would be easier just to update all syntax for Python 3 and be done with it.

umlaeute:
any news on this?

danomatika:
We are working on integrating #282. Once the multi instance stuff is working (including #343), this PR could be updated and we could add a double-precision C test project.

danomatika:
I'm just going to take this for now. The multi-instance stuff is a bit more complicated and I would rather put out a new libpd version first with this included, then has the other stuff out later.

Issue #315, opened 2020-12-01
Use libpd as a static library with CMake and MSVC


lukaspirkl:
I added the libpd as a git submodule into vendor/libpd directory. Then I added following lines into my CMakeLists.txt:

```
set(CMAKE_THREAD_LIBS_INIT "${CMAKE_SOURCE_DIR}/vendor/pthreads/lib/x86/pthreadVC2.lib" CACHE INTERNAL "" FORCE)
set(PTHREADS_INCLUDE_DIR   "${CMAKE_SOURCE_DIR}/vendor/pthreads/include" CACHE INTERNAL "" FORCE)
add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/vendor/libpd)
target_include_directories(${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/vendor/libpd/libpd_wrapper)
target_include_directories(${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/vendor/libpd/pure-data/src)
```

I added libpd_static into link libraries:

```
target_link_libraries(${PROJECT_NAME} glfw opengl32 libglew_static glm physfs-static spdlog yaml-cpp portaudio_static libpd_static)
```

For start, I added just the call to `libpd_init()` to the start of the application with the `#include "z_libpd.h"`. But I have the following linker error:

```
AudioDevice.obj : error LNK2019: unresolved external symbol __imp__libpd_init referenced in function "public: __thiscall AudioDevice::AudioDevice(unsigned int)" (??0AudioDevice@@QAE@I@Z)
```

I think there is some issue with the `EXTERN` definition in the `m_pd.h` file. Because when the lib file is compiled for the static usage the symbols don't contain the `__imp_` prefix. But when there is __declspec in the `EXTERN` it means that the linker will look for a symbol with the  `__imp_` prefix. I am not sure whether I understand the linking process correctly. So I don't know if the static linking is not supported on Windows or I am doing something wrong.

Spacechild1:
Yup, I guess we need a definition like `PD_STATIC` in `m_pd.h` which prevents the addition of `__declspec` on both ends.

Usually, libraries do it the other way around by only exporting functions if explicitly told to do so. Example: https://github.com/xiph/opus/blob/d2f6805c3f36718a742c33c56eb945c2e10646ef/include/opus_defines.h#L66. We can't do that, however, because if would affect existing build systems.

[...]

PR #316, opened 2020-12-03
Change EXTERN to EXPORT in the wrapper's z headers

myQwil:
EXPORT will be blank if the user defines the macro BUILD_STATIC

This makes it so that MinGW can link to libpd statically,
whereas before, the linker would complain about there being
undefined references.

An example of its usage has been added to samples/cpp/pdtest, which can be tested with:
    `make static`

### DIFF
--- a/libpd_wrapper/z_libpd.h
+++ b/libpd_wrapper/z_libpd.h
@@ -19,50 +19,58 @@ extern "C"
 
 #include "m_pd.h"
 
+#ifndef EXPORT
+# ifdef BUILD_STATIC
+#  define EXPORT /* Leave blank: friendly with static linking */
+# else
+#  define EXPORT EXTERN
+# endif
+#endif
+
 /* initializing pd */
 
 /// initialize libpd; it is safe to call this more than once
 /// returns 0 on success or -1 if libpd was already initialized
 /// note: sets SIGFPE handler to keep bad pd patches from crashing due to divide
 ///       by 0, set any custom handling after calling this function
-EXTERN int libpd_init(void);
+EXPORT int libpd_init(void);
 
 /// clear the libpd search path for abstractions and externals
 /// note: this is called by libpd_init()
-EXTERN void libpd_clear_search_path(void);
+EXPORT void libpd_clear_search_path(void);
 
 /// add a path to the libpd search paths
 /// relative paths are relative to the current working directory
 /// unlike desktop pd, *no* search paths are set by default (ie. extra)
-EXTERN void libpd_add_to_search_path(const char *path);
+EXPORT void libpd_add_to_search_path(const char *path);
 
 /* opening patches */
 
 /// open a patch by filename and parent dir path
 /// returns an opaque patch handle pointer or NULL on failure
-EXTERN void *libpd_openfile(const char *name, const char *dir);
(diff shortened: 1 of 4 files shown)


danomatika:
What do you think @Spacechild1? At the least, I think it the define should be `LIBPD_STATIC` or `PD_STATIC`.

danomatika:
Also note that a `STATIC` argument was recently added to the makefile, ie. `make STATIC=true`.

myQwil:
Yes, building a static library is fine. It's linking to the static library where issues start to arise, and from what I've seen so far, it's only an issue for Windows users working with MSYS2/MinGW

danomatika:
Yup, you're timing is good: #315. :)

Spacechild1:
@danomatika This PR looks good, but the issue in #315 is really about the `EXTERN` definition in `m_pd.h`, so it can't be solved in libpd.

myQwil:
I'm not sure how related this is to issue #315 but when I take the Pd API route instead by removing `__declspec` from m_pd.h and then try to link to the resulting libpd static build, I'll get errors like this:
```
In file included from ../libpd/libpd_wrapper/z_libpd.h:20,
                 from ../libpd/cpp/PdBase.hpp:16,
                 from src/PdObject.hpp:16,
                 from src/PdObject.cpp:1:
../libpd/pure-data/src/m_pd.h:440:13: error: storage size of 'sys_font' isn't known
  440 | EXTERN char sys_font[]; /* default typeface set in s_main.c */
      |             ^~~~~~~~
../libpd/pure-data/src/m_pd.h:441:13: error: storage size of 'sys_fontweight' isn't known
  441 | EXTERN char sys_fontweight[]; /* default font weight set in s_main.c */
```

Spacechild1:
> by removing __declspec from m_pd.h

Did you maybe also accidentally remove the `extern` keyword? That would explain the error message.

myQwil:
line 32 of m_pd.h looks like this:
```
#ifdef _WIN32
#ifdef PD_INTERNAL
#define EXTERN
#else
#define EXTERN
#endif /* PD_INTERNAL */
#else
#define EXTERN extern
#endif /* _WIN32 */

```

Spacechild1:
Try `#define EXTERN extern`. You must not omit the `extern` keyword, otherwise the global variable forward declarations will become definitions...

myQwil:
Okay, that does work. Now that I know that, this PR probably isn't the right approach after all

libpd_wrapper/z_libpd.h
```diff
@@ -19,50 +19,58 @@ extern "C"
 
 #include "m_pd.h"
 
+#ifndef EXPORT
+# ifdef LIBPD_STATIC
+#  define EXPORT /* Leave blank: friendly with static linking */
```
Spacechild1, line 24: this should be `#define EXPORT extern`. we still need the `extern` keyword for global variables, but we don't want to DLL export/import them.

Spacechild1:
Since we need to fix `EXTERN` in `m_pd.h` anyway, I'm not sure this PR is even necessary... After the fix, `EXTERN` will also do the right thing for libpd.

myQwil:
That's what I'm thinking as well. I'll close this and leave it up to the fix being made in Pd

PR #320, opened 2021-01-19
Update CMakeLists.txt with vars for pthread paths

mxa:
option is a bool, We need to save a path.
With this fix we can use the cmake-gui app in windows


mxa:
Before: (paths were not settable in cmake-gui because they were bools)
image


Now: (now paths may be set!)
image

danomatika:
Nice. Let me know when the PR is ready.

mxa:
I think that's it.

PR #332, opened 2021-05-15
libpd_add_to_help_path

actondev:
While working with libpd (and launching the gui), I couldn't get use the help functions.
Now, by calling `libpd_add_to_help_path("/usr/lib/puredata/doc/5.reference")` (in linux) it works

### DIFF
--- a/libpd_wrapper/z_libpd.c
+++ b/libpd_wrapper/z_libpd.c
@@ -124,6 +124,12 @@ void libpd_add_to_search_path(const char *path) {
   sys_unlock();
 }
 
+void libpd_add_to_help_path(const char *path) {
+  sys_lock();
+  STUFF->st_helppath = namelist_append(STUFF->st_helppath, path, 0);
+  sys_unlock();
+}
+
 void *libpd_openfile(const char *name, const char *dir) {
   void *retval;
   sys_lock();
--- a/libpd_wrapper/z_libpd.h
+++ b/libpd_wrapper/z_libpd.h
@@ -36,6 +36,13 @@ EXTERN void libpd_clear_search_path(void);
 /// unlike desktop pd, *no* search paths are set by default (ie. extra)
 EXTERN void libpd_add_to_search_path(const char *path);
 
+/// add a path to the libpd help paths
+/// relative paths are relative to the current working directory
+/// this is triggered when calling help from pd gui.
+/// pd will search for files named <object name>-help.pd
+/// in linux typically that'd be /usr/lib/puredata/doc/5.reference
+EXTERN void libpd_add_to_help_path(const char *path);
+
 /* opening patches */
 
 /// open a patch by filename and parent dir path


danomatika:
Yes, this is normal behavior as libpd does not set *any* search paths by default.

danomatika:
Better is probably to have the local GUIs search paths added when launching the GUI, possibly by an additional argument in the start function.

umlaeute:
> Better is probably to have the local GUIs search paths added when launching the GUI, possibly by an additional argument in the start function.

i'm not sure i understand the context (libpd vs pd). which start function?

- help-paths only make sense with a GUI
- *but* help-patches are opened by the core (this is important if you are running GUI and core on different machines)

danomatika:
>i'm not sure i understand the context (libpd vs pd). which start function?

You can start a local GUI from within libpd if a Pd installation is found and the libpd instance acts as the core. The issue is that libpd by default does not add any search or help paths. This PR proposes to add a function to add to libpd's help paths.

Normally, I'd say help paths are not needed for libpd since it's essentially an embedded, headless core. When working with the GUI, they could be which is why my thinking is more along the lines of: "optionally add some help paths from the Pd installation when launching the GUI from within libpd." If if it's to difficult or fragile to divine these paths, then this PR would also work, although it foists the path finding back onto the user.

umlaeute:
the most important question for me is: why do we need to wrap the `-helppath` flag into a function in the first place?

afaict, `libpd_add_to_search_path()` will already achieve the same thing, without having to add yet another symbol of obscure use.

at least: i don't think i've used the `-helppath` startup flag in the last decade and think that this was a mistake in the first place (though `doc/5.reference/` is typically in *my* help-path anyhow)

the only drawback i can think of, is that `libpd_add_to_search_path()` will expose all pd-files within the `doc/5.reference/` folder as abstractions.
this is probably not a problem with the help-patches themselves, but there's also a very generically named `[setctl]` abstraction (used in *slop~-help*).

danomatika:
I would prefer the behavior matched desktop Pd, if possible, which is why I suggested having to done automatically if you choose to add paths when opening the GUI. We can then do help and search paths internally.

umlaeute:
i think doing it automatically in `libpd_start_gui()` is fine with me (for whatever */* have to say on this, which is precious little).

i'm pretty sure that it is much better than adding ~~`libpd_add_to_help_path()`~~.

automatically adding the help-paths has some side-effects on the (lib)Pd-instance though (obviously there are now help-paths which haven't been there before), and i wonder whether this should be undone in `libpd_stop_gui()` (otoh, the help-paths are never going to be used without the GUI, so little harm is done here; but what happens if you start different Pd-GUIs one after the other? which help-paths will be searched for in which priority?)

danomatika:
Yeah, I feel that it should be an option when starting the GUI, possibly by an additional book argument.

We could also, as you say, store the discovered paths on start and dump them again on stop. I just wonder what the right balance between “automatic” and “manual choice” is in this case.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
> I just wonder what the right balance between “automatic” and “manual choice” is in this case.

hmm, what are the choices?
so far it only seems to be about adding the `doc/5.reference/` folder to the help-paths.
is there anything else? (libpd makes it clear, that "unlike desktop pd, *no* search paths are set by default (ie. extra)")

if not, this makes the answer about the "right balance" a binary one (add, or not).

danomatika:
Manual choice would be adding the proposed function for manually adding the to the help paths, but as I’ve written previously, I don’t think it’s necessary.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
> Manual choice would be adding the proposed function for manually adding the to the help paths, but as I’ve written previously, I don’t think it’s necessary.

seems we agree than (without having anything to say here, i think it is a bad idea to add a special function to the API for such a niche-case).

anyhow, here'  another idea:
if the `libpd_start_gui()` function is still open for changes, you could add a second "flags" argument with a (currently single; but extendable in the future) flag to add the help-path (or not).

danomatika:
I think this is actually something that should be implemented upstream in `sys_startgui()`. libpd simply wraps this call.

danomatika:
Ping @millerpuckette

danomatika:
Looking at this again, it could be a problem as IOhannes notes, when you open multi copies of the GUI and possible end up adding different help paths each time and there is no way to clear the help paths. This is also tru for the search paths, however, although maybe that should be changed? I will need to let this one sit for now.

danomatika:
> anyhow, here' another idea:
> if the `libpd_start_gui()` function is still open for changes, you could add a second "flags" argument with a (currently single; but extendable in the future) flag to add the help-path (or not).

Looking at this again, this is probably a better idea. Is there an easy way to find the paths needed and add them internally?

danomatika:
@umlaeute Looking at this again, I think this approach of a manual function is 1. simple and 2. avoids any clever-yet-fragile search patterns or timing. I don't see a huge issue with it, except to note it's really only relevant when using `libpd_start_gui()` in which case I would perhaps name it something like `libpd_add_to_gui_help_path()` and add a sister `libpd_clear_gui_help_path()`. OTOH is that distinction necessary when we can just note it in the doc comments as in this PR. I dunno. :)

actondev:
Just wanted to say hi :D 
It's been a long time now that I don't use libpd, and I find it funny getting notified now and then about this

PR #333, opened 2021-06-13
change python to python3 in Makefile

shakfu:
The Makefile in the python build section still references `python` instead of `python3`. Given that `python2` is not supported anymore shouldn't the `python3` be the default?


danomatika:
I'd agree in theory but some systems still ship with `python` (Python 2.7), ie. macOS 11. OTOH the scripts have been updated to Python 3 in general, so you are probably right...

PR #351, opened 2022-04-17
Fix the bug which prevents receiving queued pd messages on windows.

### LINKED ISSUE #350: Possible bug related to ringbuffer on windows.
I am sending a float value every two seconds to pd, echo it back and fetch for messages in the ring buffer by using `libpd_queued_receive_pd_messages`.

This works as expected in `mac` and `linux` platforms but on `windows` the bound queued hook does not call.

After further investigation the part about wrapping atomic operations for different platforms caught my eye.

```c
#if __STDC_VERSION__ >= 201112L // use stdatomic if C11 is available
#include <stdatomic.h>
#define SYNC_FETCH(ptr) atomic_fetch_or((_Atomic int *)ptr, 0)
#define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \
  atomic_compare_exchange_strong((_Atomic int *)ptr, &oldval, newval)
#else            // use platform specfics
#ifdef __APPLE__ // apple atomics
#include <libkern/OSAtomic.h>
#define SYNC_FETCH(ptr) OSAtomicOr32Barrier(0, (volatile uint32_t *)ptr)
#define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \
  OSAtomicCompareAndSwap32Barrier(oldval, newval, ptr)
#elif defined(_WIN32) || defined(_WIN64) // win api atomics
#include <windows.h>
#define SYNC_FETCH(ptr) InterlockedOr(ptr, 0)
#define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \
  InterlockedCompareExchange(ptr, oldval, newval)
#else // gcc atomics
#define SYNC_FETCH(ptr) __sync_fetch_and_or(ptr, 0)
#define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \
  __sync_val_compare_and_swap(ptr, oldval, newval)
#endif
#endif
```

In the environment which I'm testing in (x64 windows machine) ` // win api atomics` are used.

Indeed, adding this piece of code in the end of `rb_write_to_buffer` function in `ringbuffer.c` yields different results when comparing `mac` and `windows`.

```c
int rb_write_to_buffer(ring_buffer *buffer, int n, ...)
{
  if (!buffer)
    return -1;
  int write_idx = buffer->write_idx; // no need for sync in writer thread
  int available = rb_available_to_write(buffer);
  va_list args;
  va_start(args, n);
  int i;
  for (i = 0; i < n; ++i)
  {
    const char *src = va_arg(args, const char *);
    int len = va_arg(args, int);
    available -= len;
    if (len < 0 || available < 0)
      return -1;
    if (write_idx + len <= buffer->size)
    {
      memcpy(buffer->buf_ptr + write_idx, src, len);
    }
    else
    {
      int d = buffer->size - write_idx;
      memcpy(buffer->buf_ptr + write_idx, src, d);
      memcpy(buffer->buf_ptr, src + d, len - d);
    }
    write_idx = (write_idx + len) % buffer->size;
  }
  va_end(args);
  
  //------ printfs added by me
  
  printf("Are they same? %d %d\n\n", buffer->write_idx, write_idx);
  SYNC_COMPARE_AND_SWAP(&(buffer->write_idx), buffer->write_idx,
                        write_idx); // includes memory barrier
  printf("After compare and swap? %d\n\n", buffer->write_idx);
  
  //-------
  return 0;
}
```

On mac or linux: 

An example output looks like

```
Are they same? 0 56

After compare and swap? 56

Are they same? 56 120

After compare and swap? 120

Are they same? 120 192

After compare and swap? 192

Are they same? 192 256

.. and so on..
```

as expected.

On the windows platform,

```
Are they same? 0 40

After compare and swap? 0
```

the output looks like this.

Of course this results with `buffer->write_idx` staying at always 0. Because of this `libpd_queued_receive_pd_messages` returns also early because of

```
size_t available = rb_available_to_read(pd_receive_buffer);
if(!available) return;
```
and we can never receive messages.

Have you ever came across this behavior on `windows`?
Any idea about why this happens?

I've read the https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange
but it looks normal to me. 

Note: By using a non-queued hook, because we omit ring buffers, the expected values are received. 

That could be a mitigation path for my case in the worst case.

alisomay:
# Overview

This is an attempt to fix #350.

The responsible part for this bug is in `libpd_wrapper/util/ringbuffer.c` line 31.

```c
    #define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \
            InterlockedCompareExchange(ptr, oldval, newval)
```

The documentation about how the arguments named for this function is a bit confusing.
https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange

It is natural that one might think that `ExChange` is the old value and `Comperand` is the new although the case is **opposite**.

Probably they wanted to mean "Value which we will exchange the destination with".

So by changing the order of the arguments the issue goes away.

```c
    #define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \
            InterlockedCompareExchange(ptr, newval, oldval)
```

I didn't do extensive testing but doing this produced the expected behavior on windows which is aligned with the other platforms.

### DIFF
--- a/libpd_wrapper/util/ringbuffer.c
+++ b/libpd_wrapper/util/ringbuffer.c
@@ -29,7 +29,7 @@
     #include <windows.h>
     #define SYNC_FETCH(ptr) InterlockedOr(ptr, 0)
     #define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \
-            InterlockedCompareExchange(ptr, oldval, newval)
+            InterlockedCompareExchange(ptr, newval, oldval)
   #else // gcc atomics
     #define SYNC_FETCH(ptr) __sync_fetch_and_or(ptr, 0)
     #define SYNC_COMPARE_AND_SWAP(ptr, oldval, newval) \


danomatika:
Good catch. I admit to writing the macros but not testing on all platforms. Looks like I misread the documentation on the windows function.

Is the change to `.project` necessary? I think we only need to fix in the macro.

alisomay:
"Is the change to .project necessary? I think we only need to fix in the macro."
I think I didn't see that, somehow the editor which I am using should have done it. 
Let me revert that.

danomatika:
Thanks!

PR #354, opened 2022-04-26
Add a Rust section to the main README.md

alisomay:
I've been working on a project to make libpd available in the Rust ecosystem. 

Currently it is published as [libpd-rs and libpd-sys](https://crates.io/search?q=libpd) where the former is for users and the latter is the raw bindings. 

It does not support multi instances yet, but the support will be there soon also. 

You may check the platform support [here](https://github.com/alisomay/libpd-rs#support).

If you also find it relevant, I'd love to list it in the main README.md.

Thanks to @danomatika for his support along the way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,6 +384,16 @@ If you install pyaudio with `pip`, you will need to install Portaudio first. On 
 
     brew install portaudio
 
+
+Rust
+------
+
+There is currently a third party Rust wrapper for libpd called [libpd-rs](https://github.com/alisomay/libpd-rs) which is actively maintained.
+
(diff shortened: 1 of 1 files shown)


alisomay:
Here is one of the examples where it could be seen in action.

https://www.reddit.com/r/rust/comments/ubqpq1/media_nannou_and_libpdrs_playing_well_together_3/

alisomay:
I know this is probably not a priority but if there is anything preventing the merge of this I'd love to address it. 
Looking forward to your feedback.

danomatika:
No reason vor the delay, other than it was forgotten. Merging now.

PR #359, opened 2022-06-09
Adding extra headers to install command

tiagosr:
These extra header file installs add support for compiling in some externals that would not work otherwise, mainly the `cyclone` and `zexy` externals.

With this change, a libpd user can add the C source for cyclone and zexy straight into their program, without needing to go through hoops compiling a separate external - and this is useful in situations where libpd is not expected to load any precompiled externals at runtime, such as when the CPU architecture is not explicitly supported. The particular case that spawned this request was when compiling libpd for an app in Linux for a MIPS4k processor.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,7 +282,6 @@ install:
 	install -m 644 pure-data/src/m_pd.h $(includedir)/libpd
 	install -m 644 pure-data/src/m_imp.h $(includedir)/libpd
 	install -m 644 pure-data/src/g_canvas.h $(includedir)/libpd
-	install -m 644 pure-data/src/s_utf8.h $(includedir)/libpd
 	if [ -e libpd_wrapper/util/z_queued.o ]; then \
 	  install -d $(includedir)/libpd/util; \
 	  install -m 644 libpd_wrapper/util/z_print_util.h $(includedir)/libpd/util; \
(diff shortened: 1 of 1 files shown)


danomatika:
@umlaeute Which headers do we consider "public"...?

danomatika:
Judging from pure-data's [src/Makefile.am](https://github.com/pure-data/pure-data/blob/18c9695ba4de543c51c2c013b45b99b8f1c35a19/src/Makefile.am#L188), the main list is:

~~~
m_pd.h z_libpd.h z_print_util.h z_queued.h
~~~

And the following are installed to `${includedir}/pd`...
~~~
m_pd.h m_imp.h g_canvas.h g_undo.h g_all_guis.h s_stuff.h s_net.h x_vexp.h
~~~

tiagosr:
I'm contemplating adding these ones that go in `${includedir}/pd` in my pull request - let me know if it's alright to do that

umlaeute:
i **really** want to get rid of `g_all_guis.h` (as a non-private header), and i think that `x_vexp.h`  should probably removed from this list as well. @danomatika probably knows best whether `s_net.h` is useful for externals.

in any case, i would at least drop the two headers i mentioned.

`s_utf8.h` has always been a private header, if cyclone is using it (i'm pretty sure that zexy is not), then this is probably a bug on cyclone's side.

danomatika:
Thanks

PR #362, opened 2022-08-04
fixed pygame examples

myQwil:
m.process() returns a byte array and needs to be
converted to an array/tuple of short ints.

pygame_test.py:
- added a clock so that it doesn't throttle the cpu

funtest.pd:
- made it quieter to prevent clipping
- changed the x-axis formula from (x + 48) * 127 to (x * 127) + 48


myQwil:
An alternative solution might be to change this line of code:
```diff
diff --git a/python/pylibpd.i b/python/pylibpd.i
index ef47e1a..12cece9 100644
--- a/python/pylibpd.i
+++ b/python/pylibpd.i
@@ -214,7 +214,7 @@ def libpd_release():
 class PdManager:
   def __init__(self, inChannels, outChannels, sampleRate, ticks):
     self.__ticks = ticks
-    self.__outbuf = array.array('b', '\x00\x00'.encode() * outChannels * libpd_blocksize())
+    self.__outbuf = array.array('h', '\x00\x00'.encode() * outChannels * libpd_blocksize())
     libpd_compute_audio(1)
     libpd_init_audio(inChannels, outChannels, sampleRate)
   def process(self, inBuffer):

```

danomatika:
> An alternative solution might be to change this line of code instead:
> 
> ```diff
> diff --git a/python/pylibpd.i b/python/pylibpd.i
> index ef47e1a..12cece9 100644
> --- a/python/pylibpd.i
> +++ b/python/pylibpd.i
> @@ -214,7 +214,7 @@ def libpd_release():
>  class PdManager:
>    def __init__(self, inChannels, outChannels, sampleRate, ticks):
>      self.__ticks = ticks
> -    self.__outbuf = array.array('b', '\x00\x00'.encode() * outChannels * libpd_blocksize())
> +    self.__outbuf = array.array('h', '\x00\x00'.encode() * outChannels * libpd_blocksize())
>      libpd_compute_audio(1)
>      libpd_init_audio(inChannels, outChannels, sampleRate)
>    def process(self, inBuffer):
> ```

Without looking at the wrapper, I think there is still the option to use other sample formats (float, etc), so the api should returns bytes...

myQwil:
I need to close this PR and re-open it because I made the mistake of putting it in my fork's master branch.

PR #368, opened 2022-11-28
Build and run javalib correctly on M1 macs

funkyfourier:
Had to make various changes to make the jarfile build and run correctly on M1 macs. Have refactored `NativeLoader.detectSystem()` a little bit. It is hard for me to test it for all cases, but unless you are on "aarch64" it should work the same as before.

Have also adjusted `PLATFORM_CFLAGS = ...` in Makefile so that it works on my system with openjdk@11 installed and `JAVA_HOME` set correctly. It seems like this depends on what kind of Java you have installed and the position of the moon and the stars, but I **think** it will work for the most part now.


java/org/puredata/core/NativeLoader.java
```diff
@@ -48,13 +48,20 @@ public NativeLibraryLoadError(String message, Throwable cause) {
...
 
-  private static void detectSystem() {
-    osArch = System.getProperty("os.arch").toLowerCase();
-    if (osArch.indexOf("64") != -1) {
-      osArch = "x86_64";
-    } else if (osArch.indexOf("86") != -1) {
-      osArch = "x86";
+  private static String determineOsArch(){
+    final String arch = System.getProperty("os.arch").toLowerCase();
+    if (arch.indexOf("aarch64") != -1) return "arm64";
```
danomatika, line 53: Can you keep the same formatting as the previous checks?

```java
if (arch.indexOf("aarch64") != -1) {
  return "arm64";
}
```

danomatika:
I would double-check the `JAVA_HOME` change as, for instance, on my system `/System/Library/Frameworks/JavaVM.framework` exists but `JAVA_HOME` is not set in my env vars.

What do you think @nettoyeurny ?

java/org/puredata/core/NativeLoader.java
funkyfourier: Seems sensible. Also removed the superfluos `else if`.

nettoyeurny:
It seems reasonable to expect JAVA_HOME to be set if you're doing Java development, but I haven't worked with Java in ages, so take this with a grain of salt :)

funkyfourier:
Any updates on this? It still does not run on Apple Silicon macs.

danomatika:
I also haven't used Java directly in a while. I assume Java now uses dylib and JavaNativeFoundation? When did that change (curious) as I don't want to end up breaking projects. If it was recently, then we should have some way for people on older systems to at least override these settings, as required, and add info to this effect to the readme. If the change was a long time ago and nobody has really noticed before then let's go ahead and adopt the change as is. Just double-checking.

funkyfourier:
@danomatika I'm afraid I don't know the answers to what you are asking. I use macs only at work and I do not usually  keep an eye on things like these. IIRC I also had these issues (dylib/JavaNativeFoundation) on the intel Macbook Pro running OS X 11.6.1 back in 2021.

danomatika:
The diff shows a change to the fundamental lib ext and location. My question was more to know if you had changed this *and* checked if this is true on all platforms? If not, then we need to determine if it will possibly break things for non-macOS systems.

danomatika:
Thank for this

PR #369, opened 2022-12-17
fixed pygame examples

myQwil:
m.process() returns a byte array and needs to be
converted to an array/tuple of short ints.

pygame_test.py:
- added a clock so that it doesn't throttle the cpu

funtest.pd:
- made it quieter to prevent clipping
- changed the x-axis formula from (x + 48) * 127 to (x * 127) + 48

This is a continuation of PR #362


danomatika:
Looks good. Thanks!

PR #370, opened 2022-12-17
cpp: address level 3 warnings

myQwil:
PdBase.hpp:
- unused parameter 'var' - Ln [823, 987]
- comparison of integer expressions of different signedness - Ln 1075

PdReceiver.hpp & PdMidiReceiver.hpp:
- 28 unused parameter warnings

PdTypes.hpp:
- type qualifiers ignored on function return type - Ln 268


danomatika:
Thanks

PR #377, opened 2023-11-04
Obj-C native instance support

danomatika:
This PR implements prelim Obj-C native instance support.

Inspired by the C++ wrapper, a new `PdInstance` class wraps addressing separate instances when built with `PDINSTANCE` defined or when linking to libpd-ios-multi via the libpd Xcode project. If `PDINSTANCE` is not defined, each `PdInstance` address the same main libpd instance in the background.

`PdAudioUnit` now creates an instance in the background and there are init functions for both `PdAudioUnit` and `PdAudioController`. Uses of `PdBase` have been removed from both classes and the `samples/objc/ios/pdtest` example.

I am debating replacing PdBase with PdInstance but this would break existing projects, so I have held off for this prelim test. For a transition, I could imagine the following:
* PdBase becomes PdInstance by adopting the instance methods
* a new separate class extension file adds static methods for those who cannot update projects, ala `PdBase+statics.h`


danomatika:
Example basic usage:
```obj-c
PdInstance *pd = [[PdInstance alloc] init];
[pd subscribe:@"fromPD"];
[pd openAudioWithSampleRate:48000
                 inputChannels:0
                outputChannels:2];
// etc...
```

With PdAudioController, PdAudioUnit creates a PdInstance on init and the PdAudioController `instance` getter accesses it:
```obj-c
PdAudioController *audioController = [[PdAudioController alloc] init];
[audioController configurePlaybackWithSampleRate:44100
	                                                        inputChannels:1
	                                                      outputChannels:2
	                                                           inputEnabled:YES];
PdInstance *pd = audioController.instance;
...
```

Setting a specific instance when using PdAudioController:
```obj-c
PdInstance *pd = [[PdInstance alloc] init];
[pd subscribe:@"fromPD"];
PdAudioController *audioController = [[PdAudioController alloc] initWithInstance:pd];
[audioController configurePlaybackWithSampleRate:44100
                                                                inputChannels:1
                                                              outputChannels:2
                                                                   inputEnabled:YES];
...
```

danomatika:
Ping @o-g-sus @virusys

danomatika:
Ok, I got back to this for libpd 0.15 and it's more or less working now:

PdBase still largely wraps the libpd C API and PdInstance wraps the same but sets the instance before every call. A main PdInstance is lazy created when needed and should always be valid.

The only niggle is many functions will set the internal libpd instance when calling, so if you mix Obj-C and the libpd C API, you might have to juggle that. I have noted this in the updated doc comments.

The iOS pdtest is also updated with a `pdtest-multi` target which builds iwht multiple instances enabled:
https://github.com/libpd/libpd/blob/update/objc-instance/samples/objc/ios/pdtest/pdtest/ViewController.m#L93

PR #379, opened 2023-11-29
Move HAVE_ALLOCA_H flag to PLATFORM_CFLAGS since it is not available …

lhondareyte:
…on *BSD

"alloca.h" is not available on *BSD (FreeBSD, NetBSD, OpenBSD). DragonFly have this header but it's just a wrapper for compatiblity.
This patch is tested on Darwin (macOS 14), FreeBSD 14 and Ubuntu 20-04 LTS


danomatika:
Thanks for this. We were fixing a similar issue on Windows and I incorporated the changes into the current commit stream, so I cannot merge this however you get credit of course. Also, can you double-check most the makefile build and building the python wrapper on BSD?

Issue #382, opened 2023-12-24
CmakeLists.txt needs to be updated to set HAVE_ALLOCA_H, HAVE_ENDIAN, etc


danomatika:
Changes to the Pd sources moved certain platform checks out of the internal headers into the build systems. Also, we cannot link libdl on Windows or BSD and need to link winsock on Windows. The CMakeLists.txt needs to be updated to handle this.

The overview of changes is:
* macOS:
  - CFLAGS: `-DHAVE_ALLOCA_H -DHAVE_MACHINE_ENDIAN_H -D_DARWIN_C_SOURCE -DHAVE_LIBDL`
  - linker: `dl`
* Windows:
  - linker: `ws2_32`, do *not* link `dl`
* Linux: 
  - CFLAGS: `-DHAVE_ENDIAN_H -DHAVE_ALLOCA_H -DHAVE_LIBDL`
  - linker: `dl`
* BSD:
  - CFLAGS: `-DHAVE_ENDIAN_H`
  - linker: do *not* link `dl`

The `Makefile` and `python/setup.py` are updated and can be used as examples.

I don't use cmake regularly nor can I easily test on multiple platforms, so I call on those that do to update this: @eldruin @diegov @johnkeeping @mxa @clwe @jcelerier

jcelerier:
... does any supported platform by pd today *not* have these headers?
From a quick check:
- endian.h is in glibc since at least 2.9 (2008) and in musl for at least 13 years. alloca.h has been on glibc for at least 26 years.
- I see on my macOS machin/endian.h a copyright header from 1995 and libdl has been here since forever too

in any case, I can whip up a quick PR for the remaining cases (linking libdl when necessary - CMake already has CMAKE_DL_LIBS which encodes the correct per-platform logic for it)

danomatika:
> ... does any supported platform by pd today _not_ have these headers?

Some do, yes, and the desire in the Pd core is to move many older platform checks out of the headers. I don't fully agree but that's how it is for 0.54 so we need to change the CFLAGS on our end, ie. https://github.com/pure-data/pure-data/pull/2081.

> in any case, I can whip up a quick PR for the remaining cases (linking libdl when necessary - CMake already has CMAKE_DL_LIBS which encodes the correct per-platform logic for it)

Yeah, that would be helpful. The list above is just what has been added to the Makefile from feedback on other platforms, ie. Windows and BSD being broken with libod 0.14.0.

danomatika:
[...]

PR #383, opened 2023-12-26
Various cmake fixes and improvements

jcelerier:
As per #382. This updates CMake minimum version to what's available in current Debian stable as this allows to simplify the scripts quite a bit (thanks to CMake now supporting detection of LINUX, MINGW, etc directly).
This also changes to a more declarative style as far as possible.


danomatika:
Thanks. I realized when we made the updates the last days that the CMake build file was out of date.

Any time to check this on Windows @Spacechild1?

jcelerier:
I also added a CI that will test debian, ubuntu, macOS, MSVC and MinGW

jcelerier:
https://github.com/jcelerier/libpd/actions/runs/7331593386 let's see what it gives...

jcelerier:
almost there, looks like there is a remaining bug with clang/lld on MinGW (the preferred platform for mingw nowadays): https://github.com/jcelerier/libpd/actions/runs/7331686672/job/19964708554 - it seemsto choke on DLL access of

    extern __thread t_pdinstance *pd_this; 

which makes sense - this is missing dllexport / dllimport, it should be EXTERN instead of extern I assume on this line

jcelerier:
also I'll let someone else link to winpthreads on the MSVC case on the CI, I have zero desire to fight with this !

alisomay:
@jcelerier Great work!

I just came back to update [libpd-sys](https://github.com/alisomay/libpd-sys) to the latest master of libpd today. 

I have also realized that it doesn't build due to some missing files in PD_SOURCES
-  pure-data/src/g_vdial.c
-  pure-data/src/g_vslider.c
-  pure-data/src/g_hdial.c
-  pure-data/src/g_hslider.c

These files indeed do not exist in pd source. 

If this is an overlooked thing or a real issue would you include fixes in this PR also?

alisomay:
My bad, it is a sub module updating issue. Sorry for the noise. 😊

danomatika:
> also I'll let someone else link to winpthreads on the MSVC case on the CI, I have zero desire to fight with this !

For whoever wants to do this: A copy of the dlls for mingw32 and mingw64 are included in the `libs` folder.

danomatika:
> almost there, looks like there is a remaining bug with clang/lld on MinGW (the preferred platform for mingw nowadays): https://github.com/jcelerier/libpd/actions/runs/7331686672/job/19964708554 - it seems to choke on DLL access of
> 
> ```
> extern __thread t_pdinstance *pd_this; 
> ```
> 
> which makes sense - this is missing dllexport / dllimport, it should be EXTERN instead of extern I assume on this line (and EXTERN should exapnd to the correct attribute to export symbols from a DLL..., not just "extern"...)

Is this the declaration in m_pd.h? We'll have to get it fixed upstream. @umlaeute What do you think?

danomatika:
Thanks for the modernization. Let me know when y'all are happy with this and I can merge for a 0.14.2 release.

jcelerier:
Ok, MSVC builds too :) don't have anything else to add to this patch so far but please other cake user test it!

alisomay:
In which systems would you like me to test it with which options? 
I have access to 
- arm64, x86_64 windows msvc and gnu
- arm64 mac

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,189 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.distro }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        pd_multi: [ON, OFF]
+        build_type: [Debug, Release]
+        distro: ["debian:bookworm", "ubuntu:lunar", "ubuntu:mantic"]
```
umlaeute, line 17: For Debian, you should use `stable` and `testing` rather than hardcoding a release name (and I think you should test both).
Unfortunately this is not possible for Ubuntu (though I'm sure that I have some code lying around somewhere to create a build matrix with a dynamically calculated list of supported/current releases)

danomatika:
I'd prefer the CI to *also* run the Makefile, but that is a point probably made in a separate issue.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Dec 28, 2023, at 1:51 PM, umläute ***@***.***> wrote:﻿
@umlaeute commented on this pull request.



In .github/workflows/builds.yaml:
> +
+on: push
+
+jobs:
+  debian:
+    name: Debian & Ubuntu
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.distro }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        pd_multi: [ON, OFF]
+        build_type: [Debug, Release]
+        distro: ["debian:bookworm", "ubuntu:lunar", "ubuntu:mantic"]

For Debian, you should use stable and testing rather than hardcoding a release name (and I think you should test both).
Unfortunately this is not possible for Ubuntu (though I'm sure that I have some code lying around somewhere to create a build matrix with a dynamically calculated list of supported/current releases)

.github/workflows/builds.yaml
jcelerier: done

jcelerier: (although personally I've been bitten enough by platforms which change under my feet to prefer fixed names - e.g. now the build status for a given commit could change in 5 years when debian:stable refers to a new distro)

jcelerier:
I agree about putting Makefile in a separate CI (likely in a diffferent GH workflow file would make sense)

I tried a fix for the last remaining error: https://github.com/pure-data/pure-data/pull/2162

.github/workflows/builds.yaml
umlaeute: which it should. so everything is good now. thanks

CMakeLists.txt
```diff
@@ -1,22 +1,29 @@
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 3.25)
```
Spacechild1, line 1: Is there a reason why we need such a recent CMake version?

CMakeLists.txt
```diff
@@ -177,10 +209,14 @@ source_group(pdextra    FILES ${PD_EXTRA_SOURCES})
...
 source_group(libpdutils FILES ${LIBPD_UTILS_SOURCES})
 
-include_directories(libpd_wrapper)
-include_directories(pure-data/src)
-
-find_package(Threads REQUIRED)
+# set the output library name for libpd depending on the settings
+set(LIBPD_OUTPUT_NAME     pd)
+if (WIN32)
+    set(LIBPD_OUTPUT_NAME libpd)
```
Spacechild1, line 215: What is the rational for this? When I try this with MinGW, I get the following files:
```
liblibpd-multi.dll
liblibpd-multi.dll.a
liblibpd-multi-static.a
```
This does not look right...

The makefile build system produces the following files instead:
```
libpd.a (static library)
libpd.lib (import library)
pd.dll (DLL)
```

Spacechild1: Note that the Windows DLL should be called `pd.dll` (without the `lib` prefix!) because that is what externals link against.

CMakeLists.txt
```diff
@@ -177,10 +209,14 @@ source_group(pdextra    FILES ${PD_EXTRA_SOURCES})
...
+endif()
+if (PD_MULTI)
+    set(LIBPD_OUTPUT_NAME ${LIBPD_OUTPUT_NAME}-multi)
```
Spacechild1, line 218: I see your reasoning for the `-multi` suffix, but it is not canonical. The makefile build system does not do this. Also, this breaks on Windows because externals always link against `pd.dll`.

CMakeLists.txt
```diff
@@ -1,22 +1,29 @@
...
 option(PD_LOCALE "Set the LC_NUMERIC number format to the default C locale" ON)
-option(PD_BUILD_C_EXAMPLES "Builds C API example programs" OFF)
+option(PD_DEFINE_EXTERN "Define EXTERN to a custom string for special linking purposes" "")
+option(LIBPD_STATIC "Build static library" ON)
+option(LIBPD_SHARED "Build shared library" ON)
+cmake_dependent_option(PD_BUILD_C_EXAMPLES "Builds C API example programs" OFF "LIBPD_SHARED" OFF)
 
 if (MSVC)
-    set(CMAKE_THREAD_LIBS_INIT CACHE PATH "Path to pthreads library binary file (ex. C:/src/vcpkg/packages/pthreads_x64-windows/lib/pthreadVC3.lib)")
+    set(PTHREADS_LIB CACHE PATH "Path to pthreads library binary file (ex. C:/src/vcpkg/packages/pthreads_x64-windows/lib/pthreadVC3.lib)")
```
Spacechild1, line 18: I guess this should be `FILEPATH` instead of `PATH` because we need to specify a file, not a directory.

CMakeLists.txt
Spacechild1: Ah, I see! This is for MSVC. There it produces the following output:
```
libpd-multi.dll (DLL)
libpd-multi.lib (import library)
libpd-multi-static.lib (static library)
```
So I guess this should be `if (MSVC)` instead of `if (WIN32)`. (Also, the DLL itself must be called `pd.dll`)

Spacechild1:
With https://github.com/pure-data/pure-data/pull/2165 applied, libpd with `-DPD_MULTI` and `-DPD_BUILD_C_EXAMPLES` builds fine here on Windows with GCC, Clang and MSVC!

CMakeLists.txt
jcelerier: hmm I think I just followed what was there in the previous version of the CMakeLists

CMakeLists.txt
jcelerier: yes, it's to enable the LINUX, BSD variables.
Note that this is the CMake version available in the current Debian Stable.

CMakeLists.txt
jcelerier: I don't know, most of the -multi logic was pre-exisiting I just moved it a bit in the CMakeLists.txt during the refactor. Shouldn't libpd with and without MULTI have different names? As they have a different ABI, you certainly don't want linking with pd.{dll,lib,dylib} to succeed if you have a PD_MULTI lib as then you'll get runtime crashes... (and conversely, you should definitely *not* be able to easily link a non-multi external against a multi Pd

CMakeLists.txt
Spacechild1: Fair enough!

CMakeLists.txt
Spacechild1: > I don't know, most of the -multi logic was pre-exisiting I just moved it a bit in the CMakeLists.txt during the refactor

Ah, I didn't realize that, sorry!. In that case, this is of course out of scope for this PR. I will open dedicated issues.

---

> As they have a different ABI, you certainly don't want linking with pd.{dll,lib,dylib} to succeed if you have a PD_MULTI lib as then you'll get runtime crashes...

That's not true. From the perspective of an external, there is no difference between a single-instance and multi-instance Pd - except for `pd_this`, which is only exported as a workaround for the global symbols `s_float`, `s_list`, etc. If an external consistently uses `gensym`, it will run in any Pd. If it uses any of the global symbols, it won't load in the "wrong" Pd because of unresolved symbols. There are no runtime crashes *)

*) There is one specific case I know of: if an external wants to call Pd API functions from a worker thread (which is not safe anyway, because `sys_lock`/`sys_unlock` can deadlock) and does not call `pd_setinstance` when loaded in a multi-instance Pd. (In this case, `pd_this` would be `NULL` and most API function calls would crash.)

CMakeLists.txt
Spacechild1: I see! See my other comment https://github.com/libpd/libpd/pull/383#discussion_r1442268113

CMakeLists.txt
jcelerier: I don't understand. I see in m_pd.h: 

```
struct _pdinstance
{
    double pd_systime;          /* global time in Pd ticks */
    t_clock *pd_clock_setlist;  /* list of set clocks */
    t_canvas *pd_canvaslist;    /* list of all root canvases */
    struct _template *pd_templatelist;  /* list of all templates */
    int pd_instanceno;          /* ordinal number of this instance */
    t_symbol **pd_symhash;      /* symbol table hash table */
    t_instancemidi *pd_midi;    /* private stuff for x_midi.c */
    t_instanceinter *pd_inter;  /* private stuff for s_inter.c */
    t_instanceugen *pd_ugen;    /* private stuff for d_ugen.c */
    t_instancecanvas *pd_gui;   /* semi-private stuff in g_canvas.h */
    t_instancestuff *pd_stuff;  /* semi-private stuff in s_stuff.h */
    t_pd *pd_newest;            /* most recently created object */
#ifdef PDINSTANCE
    t_symbol  pd_s_pointer;
    t_symbol  pd_s_float;
    t_symbol  pd_s_symbol;
    t_symbol  pd_s_bang;
    t_symbol  pd_s_list;
    t_symbol  pd_s_anything;
    t_symbol  pd_s_signal;
    t_symbol  pd_s__N;
    t_symbol  pd_s__X;
    t_symbol  pd_s_x;
    t_symbol  pd_s_y;
    t_symbol  pd_s_;
#endif
#if PDTHREADS
    int pd_islocked;
#endif
};
```

so if anyone with -DPDINSTANCE=1 happens to access or a pdinstance object somewhere where libpd was compiled with -DPDINSTANCE=0 this will lead to immediate crash, stack overflow, undefined behaviour and all the fun stuff. e.g. just someone doing `auto instance = pd_maininstance;` would be enough.

jcelerier: okay.. anything that can be done wrong without causing a hard build error doesn't register as "safe" in my book but if it's not a problem for the project :)

CMakeLists.txt
jcelerier: pushed!

jcelerier:
okay, normally we went through all the reviews, let's wait for the build and then ACK :)
https://github.com/jcelerier/libpd/actions/runs/7478303922

jcelerier:
Build seems good, since the remaining failures are to be fixed in upstream pd

jcelerier:
I think this can be merged? :)

alisomay:
Any plans on merging? :)

danomatika:
I'll get to this when I can, busy here. Does it compile ok with the current pure-data upstream master?enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Apr 21, 2024, at 3:29 PM, Ali Somay ***@***.***> wrote:﻿
Any plans on merging? :)

jcelerier:
> Does it compile ok with the current pure-data upstream master

I updated a branch on top of master with these patches and it seems to work fine : https://github.com/jcelerier/libpd/tree/ossia-2024-07-07

jcelerier:
yay :)))

danomatika:
Note: `master` no longer builds libpd C sources from `libpd_wrapper`. I updated CMakeLists.txt to reflect that but have not tested.

danomatika:
Can someone check this? The CI Windows builds fail.

danomatika:
> Note: `master` no longer builds libpd C sources from `libpd_wrapper`. I updated CMakeLists.txt to reflect that but have not tested.

Oh and I reverted this change for now. It lives in a branch again...

jcelerier:
from the error it seems to be a missing msvc flag to enable C11 atomics: 

    C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\vcruntime_c11_stdatomic.h(36,24): error C2061: syntax error: identifier 'atomic_bool' [D:\a\libpd\libpd\build\libpd.vcxproj]
      (compiling source file '../pure-data/src/d_fft_fftsg.c')

What's the branch ? I can look into it

danomatika:
It's in `master`.

jcelerier:
ah ok, I thought the entire branch had been reverted

danomatika:
EH, I merged it then made additional commits, so it got messy. I only reverted the two main commits for now. I will likely have to redo those commits in a newer clean branch.

For this issue, I probably broken something. ;P

PR #400, opened 2024-06-16
Add a Zig example

myQwil:
Currently has only been tested on Linux.
For the desktop version, run the command:
 `zig build --release=fast run`

To run the example in a web browser, first install emsdk. Once emsdk is installed, set it up by running
`emsdk install latest`
Find the folder where it's installed and run
`zig build --release=fast -Dtarget=wasm32-emscripten --sysroot [path to emsdk]/upstream/emscripten run`


danomatika:
Honestly, this is beyond the scope of examples here. I think it would be better in a separate repo where someone else can maintain it.

myQwil:
I've recently moved the module to its own repo. What I could do is move the build portion to another separate repo, and then the example could just grab those repos as dependencies, so it would basically just be main.zig and test.pd, much smaller compared to its current state.

danomatika:
I have not used Zig, is a "module" basically a wrapper or bindings for libpd? If so, then yes that makes the most sense. If you you have a some simple examples, then it makes sense to include them there, otherwise a large-scale example set is best in a separate repo.

myQwil:
> is a "module" basically a wrapper or bindings for libpd?

A module is basically any zig source code. In our case, it's being used as a wrapper. It exposes all of the libpd functions, which can then be imported into a zig project, but it also has the library itself attached to it.

myQwil:
This example has been added to the zig module. That's probably the more appropriate place to put it.

Module can be found here: https://github.com/myQwil/libpd_zig

PR #402, opened 2024-11-18
ci: try to fix MSVC atomic issue

jcelerier:


jcelerier:
if possible @danomatika I'd like to get access to CI for this branch to be able to test, likely it's not going to only take one try..

danomatika:
Merging to `develop` which is up to date with `master`. Also, how do I give you access to the CI?

jcelerier:
either here : 
image
you should have a button to allow the workflow to run

or in the settings: 

image

jcelerier:
ahh wait I think I have to also add something to the workflow

jcelerier:
ah it's good it was on my side, my bad ! :)

jcelerier:
all good @danomatika - it's mainly that pd requires C11 support now

danomatika:
Nice, thanks. Right, C11 is needed for the atomics now used internally and also by z_ringbuffer which previously used OS-specifics if C11 wasn't available.

Spacechild1:
> Right, C11 is needed for the atomics 

Pd does *not* require C11! It will try to use C11 atomics if available, otherwise it fallbacks to platform-specific implementations. See: https://github.com/pure-data/pure-data/blob/68a30a3e2672d0d57baff7d862d6b611c79d53ec/src/m_private_utils.h#L18-L72

Strangely, I'm pretty sure I've tested this with Visual Studio 2022, at least with the old MSVC makefile in the Pd repo.

Anyway, enforcing C11 is certainly not the correct solution here.

See https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/

> Atomics are available in Visual Studio 2022 version 17.5 Preview 2 with the /experimental:c11atomics flag, in /std:c11 mode or later. 

If I read this correctly, we don't need to set `/experimental:c11atomics` if we are building in `/std:c11` mode. I certainly can't remember having to set this flag...

The problem is that MSVC only added stdatomic in 17.5 Preview 2, *after* they started to officially support C11. This means we can't just test for C11 with `__STDC_VERSION__`, but we must also check the MSVC version. Actually, I've just discovered an issue with the checks in `m_private_utils.h`: we do check for VS 22 17.5, but we don't check for C11! This obviously fails when 17.5 runs at older C++ versions.

I *think* we just have to replace

```
#ifdef _MSC_VER
/* NB: Visual Studio only added C11 atomics support in version 17.5 */
#if _MSC_VER >= 1935
#define HAVE_C11_ATOMICS
#endif
#elif __STDC_VERSION__ >= 201112L
#define HAVE_C11_ATOMICS
#endif
```

with 

```
#ifdef _MSC_VER
/* NB: Visual Studio only added C11 atomics support in version 17.5;
__STDC_VERSION__ is only defined starting from C11 */
#if defined(__STDC_VERSION__) && _MSC_VER >= 1935
#define HAVE_C11_ATOMICS
#endif
#elif __STDC_VERSION__ >= 201112L
#define HAVE_C11_ATOMICS
#endif
```

Can someone try this?

PR #404, opened 2024-11-19
transition building libpd C sources from libpd_wrapper to pure-data/src

danomatika:
This PR transitions building the libpd C sources from the included `libpd_wrapper` directory to those included in `pure-data/src`. As these sources are largely stable, the idea is to remove duplication however at the expense of requiring changes or bugfixes for C to be submitted to the upstream `pure-data` repo itself. A further step would be to transition to possibly using the build stems in pure-data itself for libpd but this could be possibly done at another step.


danomatika:
My thought was to make the transition gentler but perhaps you are right.

Spacechild1:
> As for the basic build system, one option is to turn the current makefile into a wrapper which just calls the pure-data/libpd/Makefile however I think there should be some guarantees about this not being broken arbitrarily.

This could be a nice strategy for a gradual transition. However, if you want to able to iterate quickly, it makes more sense to keep the makefiles separate. Instead I would suggest to harmonize `pure-data`'s makefile with ours, so that we can simply contribute any changes back to `pure-data`. Basically, we would treat the Makefile just as we treat the `z_*` sources.

danomatika:
They can't be *exactly* the same since the source layout is different, but maybe some path checking can enable this?

Spacechild1:
Side note: we might also rethink Pd's project structure a bit. Does it really make sense to have the libpd C wrapper files in `src` while the `libpd` folder only contains the makefiles? I would rather have both the makefile and the sources in the `libpd` folder. The folder itself could be moved into `src` - or stay where it is.

PR #421, opened 2025-11-06
Update .NET wrapper

### LINKED ISSUE #417: libpdcsharp naudio example throwing exceptions
I am attempting to use libpdcsharp in a console application. I have copied the implementation of NewApiPdProvider.cs and Program.cs from the C# example, but I am getting this error:

`System.AccessViolationException: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.'`
When it's running the RefillBuffer():

```
void RefillBuffer()
{
    while (_circularBuffer.Count < _minBuffer)
    {
        // Compute audio. Take care of the array sizes for audio in and out.z
        _pd.Process(Ticks, new float[0], _pdBuffer);
        _circularBuffer.Write(PcmFromFloat(_pdBuffer), 0, _pdBuffer.Length * 4);
    }
}
```

Steps taken:

I have installed the libpd nuget and copied both the libpdcsharp.dll and libwinpthread-1.dll to my project root and set them to copy over. I have also copied the test.pd, updated the loadpatch path and set it to copy over as well.

Originally I was getting this error:

`System.BadImageFormatException: 'An attempt was made to load a program with an incorrect format. (0x8007000B)'`

So I changed the target from Any CPU to x86, even though the docs says it should work with an Any CPU build even on a 64-bit CPU.

Does anyone know what I am doing wrong? I have been at this for hours now and non the wiser.

residuum:
This is an update to the .NET wrapper:

- Fixes #417
- The library is now working with .NET Core as well as .NET Framework
- NuGet package has been update
- Update in Readme for current MSYS2, and where to get the obsolete 32bit version
- Explanation in Readme about cross-compilation for other platforms on Linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,25 @@ Once libpd has built successfully, the compiled libraries will be found in the `
 
 Install the core build requirements using your distribution's package manager. For Debian, you can install the compiler chain, autotools, & gettext with:
 
-    sudo apt-get install build-essentials
+    sudo apt install build-essentials
+
+If you want to cross-compile for other architectures, you need additional packages and set the `CC` environment variable for the architecture. Examples using Debian based amd64 distribution:
+
+For 32bit Intel:
+    
+    sudo apt install gcc-i686-linux-gnu
+    make CC=i686-linux-gnu-gcc
+
+For 32bit ARM:
+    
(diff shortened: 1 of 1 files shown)


danomatika:
Looks good apart form a small typo.

PR #422, opened 2025-12-10
Android Oboe support

### LINKED ISSUE #284: Replace OpenSL with Oboe for Android
# Problem

For Android, libpd uses OpenSL through [opensl_stream](https://github.com/nettoyeurny/opensl_stream). It is observed that there are several device specific issues (e.g. [Known Issues](https://github.com/nettoyeurny/opensl_stream#known-issues) and #245).

For best audio performance on Android, we have to use AAudio on API 27+ and OpenSL ES only for API 16+. We also require some device specific workarounds. Workarounds or fixes are required even for bugs in the platform as android fragmentation is still a problem and still a significant number of users are on older versions of android.

# Oboe

Oboe provides a unified interface which automatically chooses AAudio or OpenSL whichever is appropriate, includes device specific workarounds for glitches, and attempts to minimize buffer underruns. It is also being developed actively.

Google released [oboe](https://github.com/google/oboe) which hit stable release an year ago. It is also becoming [popular](https://github.com/google/oboe/blob/master/docs/AppsUsingOboe.md).

We should consider replacing the opensl glue in libpd with oboe glue which may also help address the audio problems plaguing android apps using libpd and pd-for-android.

Ant1r:
This implements the native interface for the Oboe audio driver library.

It takes care of the buffer size conversion between Pd and Oboe, separately for input and output (since Oboe can provide less input samples than output ones, mainly on start).

It also extends the PdBase jni a bit, to allow selecting input and output audio devices (which Oboe offers), and setting the audio buffer size.

The new `pdnativeoboe.so` is now loaded by PdBase loader when on Android, replacing the previous `pdnativeopensl.so`.

Closes #284.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -210,3 +210,17 @@ LOCAL_SHARED_LIBRARIES := pd
 
 include $(BUILD_SHARED_LIBRARY)
 
+
+# Build Oboe JNI binary 
+# (must be the last one because of the prefab import-module)
+
+include $(CLEAR_VARS)
+
+LOCAL_MODULE := pdnativeoboe
+LOCAL_C_INCLUDES := $(PD_C_INCLUDES) $(LOCAL_PATH)/jni
+LOCAL_SRC_FILES := jni/oboe/z_jni_oboe_shared.c jni/oboe/z_jni_oboe.cpp jni/oboe/OboeEngine.cpp
+LOCAL_SHARED_LIBRARIES := pd oboe
+LOCAL_LDLIBS := -llog
+include $(BUILD_SHARED_LIBRARY)
+
+$(call import-module, prefab/oboe)
--- a/java/org/puredata/core/PdBase.java
+++ b/java/org/puredata/core/PdBase.java
@@ -175,6 +175,27 @@ public final class PdBase {
   public native static int suggestOutputChannels();
 
   /**
+   * Select the audio input device
+   *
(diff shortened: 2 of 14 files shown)


joebowbeer (approved):
LGTM

Ant1r:
yes, normal java still builds (on my machine).
I have added 3 new methods to PdBase API, but I filled empty corresponding functions in the various jni sources.

Of course, this is a breaking change for external users, if they use their own jni: they will have to add these new functions too, unless the app won't link.

satoruf1:
Wow. This looks great! Unfortunately, due to family matters, I can't look into it in detail right now, but as soon as I have time, I'll definitely try it myself. Thank you so much!

joebowbeer (approved):
LGTM

satoruf1:
When do we plan to merge this?

danomatika:
It would be good to have someone who uses libpd with Java desktop to check. I haven't done a Java build in a long time.

danomatika:
I don't think you need to change your approach, I just wanted to know that *someone* had checked that the desktop Java still builds ok. Are the new functions applicable to Android only? If so, why not wrap them in defines etc so they are available on Android only. If accepting new key/value controls works when opening audio, perhaps that's nicer. I defer to those using the Java API to decide but I always prefer if whatever is changed tries to stay consistent with the current design, naming, usage etc.

Ant1r:
closing this in favor of #423

PR #423, opened 2025-12-30
Android Oboe support (2)

### LINKED ISSUE #284: Replace OpenSL with Oboe for Android

Ant1r:
This implements the native interface for the Oboe audio driver library.

It takes care of the buffer size conversion between Pd and Oboe, separately for input and output (since Oboe can provide less input samples than output ones, mainly on start).

The new `pdnativeoboe.so` is now loaded by PdBase loader when on Android, replacing the previous `pdnativeopensl.so`.

Closes #284.

This PR is similar to #422, but involves less change in the Java API.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -210,3 +210,17 @@ LOCAL_SHARED_LIBRARIES := pd
 
 include $(BUILD_SHARED_LIBRARY)
 
+
+# Build Oboe JNI binary
+# (must be the last one because of the prefab import-module)
+
+include $(CLEAR_VARS)
+
+LOCAL_MODULE := pdnativeoboe
+LOCAL_C_INCLUDES := $(PD_C_INCLUDES) $(LOCAL_PATH)/jni
+LOCAL_SRC_FILES := jni/oboe/z_jni_oboe_shared.c jni/oboe/z_jni_oboe.cpp jni/oboe/OboeEngine.cpp
+LOCAL_SHARED_LIBRARIES := pd oboe
+LOCAL_LDLIBS := -llog
+include $(BUILD_SHARED_LIBRARY)
+
+$(call import-module, prefab/oboe)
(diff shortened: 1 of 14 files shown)


joebowbeer:
Why is this a draft?

Do you want to (first) create a PR in pd-for-android that uses this commit?

Ant1r:
So I've added a getter for runtime info from the audio implementation. With Oboe, this can be used to detect xruns, or query the latency, for example for abl_link fine tuning purpose.

The PR for pd-for-android is ready (https://github.com/libpd/pd-for-android/pull/173).

I think that's enough for me now, ready for review!

danomatika:
Looks good to me. The only detail I can think of is perhaps naming "inputDeviceId" / "outputDeviceId" without the "id" naming ala "inputDevice" *but* I assume this is a convention used by Oboe / Android and, if so, let's keep it.

joebowbeer:
Re. naming, if the single parameter is an 'id' then the method name should be an 'id'

I don't see any red flags, so LGTM.

I do wish there were more automated checks in this repo: build everything incl. Java, lint, unit test, smoke test (e2e). That would give me more confidence as a reviewer. I also suggest protecting the master branch.

